PR #4, opened 2015-07-10
feat(all) : add initial implementation

behzad888:
Hi, [gulp build] not worked in the babel-dts-generator , so i dont change the build.js file.
please check that.
thank you.


EisenbergEffect:
Thanks for the contribution. Unfortunately, we can't take a jQuery dependency in this library. We've got several code-bases internally that we are merging and will have an initial implementation up soon. Then we can iterate on it. I'm sure you'll have some good ideas for that.

PR #5, opened 2015-07-12
feat(all) : add initial implementation

behzad888:
Do not build with gulp , please check .d.ts generator in build.js file


EisenbergEffect:
We have three implementations already from the core team. We are in the process of combining them into a single standard plugin. After that code is pushed would be a good time to contribute to this repo.

PR #7, opened 2015-07-28
Fix build and overlay position

michaelmalonenz:
@PWKad Seeing as you're in the mood for merging... I have a couple of slight fixes that might interest you.  They might be things you've picked up on your own, but the build fix is really annoying - took me a wee while to figure out what was happening.


PWKad:
Just got finished merging and re-basing so after I push it out here in a minute I'll ask you to rebase these changes on top :)

michaelmalonenz:
ok, sounds good.

EisenbergEffect:
Merged.

PR #8, opened 2015-07-28
feat(dialog): Add tests and usage examples

PWKad:
Shows how to use the dialog (prompt and custom)
Add unit tests


EisenbergEffect:
Hmm. It looks like this doesn't have the latest code I sent. For example, I changed up the controller so that it would not reject promises on cancel. After trying this out in a real app, it proved to be problematic because in many cases, the developer doesn't care about the cancel case and so doesn't register a catch block. This results in errors in the console. We'll get this merged in and then I may make some changes based on that.

EisenbergEffect:
Ok. This was merged...but it was kind of strange the way the branching worked out. In any case, there is one broken test "the Dialog Renderer calls the corresponding controller methods"

PWKad:
Yeah I don't know what code was missed but if you can send me the updated version I'll diff it.  I'll squash the commits before merging to clean up history.

PR #11, opened 2015-07-30
fix(dialog-renderer)

michaelmalonenz:
attached the centerHorizontalOnly setting to the horizontal alignment
added the setting to the defaults for clarity
updated the docs to show the settings

@PWKad @EisenbergEffect Corrected the horizontal setting, and added some documentation around the settings.


EisenbergEffect:
Thanks!

PR #17, opened 2015-08-12
fix: use keyup for escape key handler

jbellsey:
Keyup is a more browser-compatible event for handling the escape key. Keydown is also viable, if you prefer to handle it that way.

The keypress event is not firing in Chrome (OSX v43).


Aaike:
Thanks for the contribution !

PR #21, opened 2015-08-25
chore(build):

michaelmalonenz:
update the package versions of the test dependencies to work with the systemjs loader 0.18 and get the tests running

@PWKad I've removed a failing assertion, because it no longer makes sense, but I couldn't see an obvious replacement (on account of the model being less exposed now), so I'll leave that bit to you.  Apologies for not running the tests prior to my original PR!


PWKad:
Thanks for resolving this!

michaelmalonenz:
No worries :)

PR #36, opened 2015-09-22
feat(scrolling): Prevent body scrolling and allow to scroll only dial…

### LINKED ISSUE #31: Scrolling
Bootstrap modal is setting `.modal-open` class on the `body` element to set `overflow: hidden`. Can you add this behaviour so that scrolling will not scroll page under dialog?

Mordred:
…og itself

This should fix #31


PWKad:
@EisenbergEffect CLA check

EisenbergEffect:
@PWKad Confirmed. You are good to go.

Issue #74, opened 2015-12-09
"settings.lock = false" does not close dialog when clicking ouside dialog


jc1001:
Hi there,

According to the readme, when the user clicks out of the dialog, it will close and resolve with a cancel.  I can't get this functionality to work (I did confirm that the lock value on the settings object is set to false).  

After some digging in dialog-renderer.js (line 66), I found that there is a click event on the modal overlay that will call the cancel method on the controller when lock is set to false.  The problem is that the modal overlay is rendered behind the modal container (both fill the whole browser window in my application).  From the code (dialog-renderer.js - line 51) I can see that the modal container has a higher z-index than the modal overlay (which basically prevents the click to reach the overlay).  When manually adjusting these values in the browser diagnostic tools, I can get the click event to trigger and close the dialog (basically increased the z-index of the overlay - just as a test).

Markusxmr:
:+1:

Markusxmr:
:+1:

PWKad:
@jc1001 If your dialog is covering the entire window there is no area to click on outside of the dialog correct?  Unless I am mis-understanding your explanation it sounds like that is the case.

Perhaps if you could reproduce the issue and let me know steps to recreate the problem I can help take a look if it is indeed an issue or the intended behavior.

jc1001:
Hi there,

Thanks for getting back to me so soon.  I reproduced the issue by doing the following:
* Downloaded the ES2016 starter kit from your site and ran it with the NodeJS option (http-server -o)
* Installed the aurelia-dialog with jspm and added it as a plugin
* Added the edit-person dialog as outlined on the dialog readme.

When I run the sample, I get the following screen:
![dialog1](https://cloud.githubusercontent.com/assets/15125812/11743703/445921f2-a012-11e5-875a-cb47c29562bf.png)

When I use the browser diagnostic tools, I see the following - notice that the highlighting for ai-dialog-overlay and ai-dialog-container both fill the whole screen:
![dialog2](https://cloud.githubusercontent.com/assets/15125812/11743778/e65b91ba-a012-11e5-880f-502914391cd2.png)
![dialog3](https://cloud.githubusercontent.com/assets/15125812/11743780/ea0807da-a012-11e5-9f37-910c6f7315be.png)

Since the click event is bound to the overlay (which has a lower z-index), it is not picked up when clicking outside the visible part of the dialog.

I've attached the solution that I've tested on.
[...]

PR #79, opened 2016-01-04
disable pointer events on ai-dialog-container

benib:
This disables pointer events on ai-dialog-container to enable the close click on ai-dialog-overlay.

without this css, the event listener here https://github.com/aurelia/dialog/blob/master/src/dialog-renderer.js#L66 never catches an event as the ai-dialog-container is on top of ai-dialog-overlay.

This solution only works for browsers supporting `pointer-events`: http://caniuse.com/#search=pointer-events

### DIFF
--- a/src/styles/dialog.less
+++ b/src/styles/dialog.less
@@ -23,6 +23,9 @@ ai-dialog-container {
   overflow-x: hidden;
   overflow-y: auto;
 
+  // prevent pointer events to enable click event on ai-dialog-overlay
+  pointer-events: none;
+
   -webkit-overflow-scrolling: touch;
   // Prevent Chrome on Windows from adding a focus outline. For details, see
   // https://github.com/twbs/bootstrap/pull/10951
@@ -54,6 +57,8 @@ ai-dialog {
   border-image-outset: initial;
   border-image-repeat: initial;
   background: white;
+
+  pointer-events: initial;
 
   > ai-dialog-header {
     display: block;


benib:
Fixes #74 for browsers supporting `pointer-events`.

EisenbergEffect:
@joelcoxokc Can you comment on this? Didn't we have some major updates to the dialog we were planning based on AI? Were those updates going to solve this issue?

joelcoxokc:
Yes, this will be resolved. The new logic also provides hooks for changing things like this.

benib:
I am fine with closing this PR if it gets resolved differently in the future. This is something users of aurelia-dialog can relatively easy fix in their own css. It's just that one need to know it. I already wanted to implement the `close on outside of modal click` thing when I found the code that is supposed to do it.

PWKad:
I think this is valid for now as it closes an open issue and only has changes to the styles (which will be replaced eventually)

@benib Thanks again for the contribution and repro.  We do have contribution requirements for a few reasons -

1. Please make sure to sign the CLA
2. Can you fix the commit message to follow the pattern `fix(overlay): disable pointer events on ai-dialog-container` so that our changelog can be generated properly?
3. If it's not too much to ask also please drop the `dist` files (we only generate them on release)

If you can take care of the first item but don't want to do the other two let me know and I can clean it up, it just messes with the PR where it looks like I did the change instead of you.

benib:
@PWKad Alright, there is a new commit in this PR following 2. and 3.
I already signed the CLA some months ago.

Please let me know if there is anything else I should do.

PWKad:
@benib  Perfect thanks!!!!

@EisenbergEffect CLA confirmation request

EisenbergEffect:
CLA is confirmed :)

timfish:
Is this in the docs? I spent a while trying to work out why I couldn't click on my custom dialog and I'm sure it was working before. When I finally worked it out, my initial attempt was to add `pointer-events: all;` to `ai-dialog-container` as that appeared to fix it, only to find that when the dialog closed it disabled `pointer-events` from my whole app! 

Am I right in saying that my initial containing element  in the view has to have `pointer-events: all;` set on it? That appears to have fixed it...

PWKad:
@timfish I think you are right that this was the cause of the issue in IE, I am testing it out now and if I switch `initial` to `all` for the `ai-dialog` it appears to fix it...  I'm going to PR this fix in, when you have a chance is there anyway you can verify?

timfish:
@PWKad Yes, just tag the fix in this issue and I'll have a look!

joelcoxokc:
Pointer-events: auto;

Instead of all.

May prevent further issues down the road.

RWOverdijk:
This is causing issues for me. I'm applying the styles from bootstrap to the modals create with this module, except I can't click my input fields anymore without closing the modal, because every click is registered on the backdrop.

PR #80, opened 2016-01-08
fix(dialog): use CSS auto margins instead of JS

### LINKED ISSUE #55: Dialog does not move when screen is resized.
The dialog does not reposition itself when the screen is resized.

jedd-ahyoung:
The dialog did not auto-center on window resize. A CSS auto margin was used, but was being overwritten by `dialogController.centerDialog()`. Removing the lines of code setting an explicit margin-left inline style on the dialog container allows CSS to auto-center the modal even upon resize.

Fixes #55.

### DIFF
--- a/src/dialog-renderer.js
+++ b/src/dialog-renderer.js
@@ -118,9 +118,6 @@ export class DialogRenderer {
     dialogController.centerDialog = () => {
       let child = modalContainer.children[0];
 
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
@PWKad Feel free to merge at will.

PWKad:
@jedd-ahyoung Verified locally, very nice thanks for the fix!

Issue #94, opened 2016-02-12
the pointer-events has been disabled after the 0.5.3 release


abbasmhd:
the `ai-dialog-container` is set to `pointer-events: none;` which is preventing all the events on children buttons 
setting `pointer-events: auto;` fixed the problem with was preventing the dialoge from closing
can you please fix this issue

PWKad:
Thanks @abbasmhd this should be fixed already in master and released = https://github.com/aurelia/dialog/commit/2c4cfa635e178071cd8080384cfc1592cbfbaef2

That commit should have done the same.  Can you verify?

abbasmhd:
thanks @PWKad  
that should do it

PR #95, opened 2016-02-12
fix(dialog): allow response to click events

abbasmhd:
fix: (issue #94) change dialog.less 

set the `pointer-events: none` to the overlay layer
prevent chrome on windows from adding a focus outline

### DIFF
--- a/src/styles/dialog.less
+++ b/src/styles/dialog.less
@@ -6,6 +6,9 @@ ai-dialog-overlay {
   right: 0;
   opacity: 0;
 
+  // prevent pointer events to enable click event on ai-dialog-overlay
+  pointer-events: none;
+
   &.active {
     opacity: 1;
   }
@@ -23,9 +26,6 @@ ai-dialog-container {
   overflow-x: hidden;
   overflow-y: auto;
 
-  // prevent pointer events to enable click event on ai-dialog-overlay
-  pointer-events: none;
-
   -webkit-overflow-scrolling: touch;
   // Prevent Chrome on Windows from adding a focus outline. For details, see
   // https://github.com/twbs/bootstrap/pull/10951
@@ -33,6 +33,12 @@ ai-dialog-container {
 
   &.active {
     opacity: 1;
+  }
+  
+  & > div {
+    // Prevent Chrome on Windows from adding a focus outline. For details, see
+    // https://github.com/twbs/bootstrap/pull/10951
+    outline: 0;
   }
 }
 
--- a/src/styles/dialog.less
+++ b/src/styles/dialog.less
@@ -25,17 +25,13 @@ ai-dialog-container {
   opacity: 0;
   overflow-x: hidden;
   overflow-y: auto;
-
   -webkit-overflow-scrolling: touch;
-  // Prevent Chrome on Windows from adding a focus outline. For details, see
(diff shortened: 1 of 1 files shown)


PWKad:
Hi @abbasmhd thanks for this contribution!

Few notes - we have contributing guidelines which show how to write commit messages so they show up in our change log properly.

Also please make sure to sign the CLA.

As I mentioned in the issue I don't think this is still a problem but I did notice there was another change to the less to not show an outline - if you can open an issue with how to reproduce that and clean up the commit we can merge this in.

Thanks!

EisenbergEffect:
@abbasmhd Any interest in following up on this?

abbasmhd:
love too follow up but i can't find the guidelines anywhere in this repo or on your website
if you can provide some links please

EisenbergEffect:
@abbasmhd There's a CONTRIBUTING.md file in every repo that will explain everything. Here's the link: https://github.com/aurelia/dialog/blob/master/CONTRIBUTING.md

abbasmhd:
@EisenbergEffect thanks Rob for the link the CLA had been signed

PWKad:
@abbasmhd If you can fix the commit message I can merge this in.

src/styles/dialog.less
```diff
@@ -34,6 +34,12 @@ ai-dialog-container {
   &.active {
     opacity: 1;
   }
+  
+  & > div {
+    // Prevent Chrome on Windows from adding a focus outline. For details, see
+    // https://github.com/twbs/bootstrap/pull/10951
+    outline: 0;
```
Gheoan: Is this necessary? It is already applied on the parent (line 32).

abbasmhd: @Gheoan yes it is necessary
the parent one it's not i just put it for extra measure
2016-02-26_13h54_24

Gheoan: @abbasmhd Fine then, however the rules should be merged like so:
```less
&, // now, this rule applies to the parent too, so remove it from above
& > div {
  // Prevent Chrome on Windows from adding a focus outline. For details, see
  // https://github.com/twbs/bootstrap/pull/10951
  outline: 0;
}
```

abbasmhd: @Gheoan the rules has been merged together

iveetoo:
@PWKad Hi, according to the docs (and the source), if `setting.lock = false` and when the user clicks outside of the dialog, it will close it and resolve it with cancel. But, when I click outside - nothing happens.

The event handler, for closing the dialog, is set on the overlay layer (https://github.com/aurelia/dialog/blob/master/src/dialog-renderer.js#L80) and setting `pointer-events: none` to `.ai-dialog-overlay` prevents all clicks on this element => the event handler is not called.

I found this issue #74, which describes the same problem about closing the dialog, but the problem were that the container was on top of the overlay and it was fixed with disabling pointer events on the `.ai-dialog-container` to enable click events on `.ai-dialog-overlay` #79

When I remove the css rule `pointer-events:none;` from `.ai-dialog-overlay` and set it on `.ai-dialog-container` the event handler is called and the dialog is closed correctly. And I haven't seen any focus outline. 

So.. Am I missing something? Is this change necessary?

PR #98, opened 2016-02-23
Attach focus value

vuorinem:
Use the value of the attach-focus attribute to decide if the element is to be focused on attach. This allows using binding to alter the focused element.


EisenbergEffect:
@PWKad Please review.

PWKad:
@vuorinem nailed it thanks for the test also like we talked about.  @EisenbergEffect can you confirm CLA please?

EisenbergEffect:
@PWKad CLA is confirmed.

PR #111, opened 2016-03-10
fix(dialog): put back centerDialog

derekpitt:


derekpitt:
@PWKad here you go! thanks man!

src/dialog-renderer.js
```diff
@@ -124,6 +122,11 @@ export class DialogRenderer {
       });
     };
 
+    dialogController.centerDialog = () => {
```
PWKad: Let's just `dialogController.centerDialog = centerDialog` here so it uses hte same function.

PWKad: Nevermind I see why you can't do that, looks good thanks.

PR #115, opened 2016-03-13
minor refactorings

mbroadst:
* completed promise cleanup that recently made it in
* moved dialog utilities to their own file with some documentation


mbroadst:
this actually had a bug in `getViewModel` where `this.compositionEngine` was being used without providing context for `this`. It wasn't picked up by the tests, so that's probably worth investigating..

PWKad:
Hi @mbroadst thanks for this contribution - 

We need to add some tests to cover new functionality at least, I'll try to go through now and give some pointers but I haven't had a chance to pull this down yet to confirm just going off of a code review.

src/dialog-controller.js
```diff
@@ -31,19 +29,16 @@ export class DialogController {
...
-            return this._renderer.hideDialog(this);
-          }).then(() => {
-            return this._renderer.destroyDialogHost(this);
-          }).then(() => {
             this.controller.unbind();
             this._resolve(returnResult);
-          });
-      }
-    });
+          })
```
PWKad: I think we are missing a semi-colon here?

src/dialog-utilities.js
```diff
@@ -0,0 +1,39 @@
+import {Origin} from 'aurelia-metadata';
+
+/**
+* Returns a view model based on a given instruction
+*
+* @param {Object} instruction the instruction from which the view model is determined
+* @return {Promise} the view model
+*/
+export function getViewModel(instruction: any, compositionEngine: CompositionEngine) {
```
PWKad: Need to ensure that this previously private method is covered by tests

src/dialog-utilities.js
```diff
@@ -0,0 +1,39 @@
...
+* @param {Object} [model]   the optional model to invoke the event with
+* @return {Promise} the result of the invocation
+*/
+export function invokeLifecycle(instance: any, name: string, model: any) {
+  if (typeof instance[name] !== 'function') {
+    return Promise.resolve(true);
+  }
+
+  let result = instance[name](model);
+  if (result instanceof Promise) return result;
```
PWKad: This shouldn't pass `gulp lint`

src/dialog-controller.js
mbroadst: it's technically not, though I can see how this is confusing. Since I'm using big arrow notation's "compact" mode (I don't know if there's a proper name for this), this whole statement is the "return" of the anon function. The reason I left the paren below this unindented was to show which `then` it was aligned with.

src/dialog-utilities.js
mbroadst:
```
mbroadst@retinoid:dialog (minor-refactorings)$ gulp lint
[14:01:23] Using gulpfile ~/Development/aurelia/dialog/gulpfile.js
[14:01:23] Starting 'lint'...
[14:01:23] Finished 'lint' after 846 ms
```

I believe your eslint rules allow single line returns, but require braces if I were to put it a line below

src/dialog-utilities.js
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+* Invokes a lifecycle event on a given instance
+*
+* @param {Object} instance  the instance to invoke the event on
+* @param {String} name      the event to invoke
+* @param {Object} [model]   the optional model to invoke the event with
+* @return {Promise} the result of the invocation
+*/
+export function invokeLifecycle(instance: any, name: string, model: any) {
```
PWKad: `invokeLifecycle` now seems to have different logic so we need tests to cover the fix so we know the expected results.  ex - 

```
describe('.invokeLifecycle', () => {
  it('should return a resolved promise if blah blah', () => {
    let result = invokeLifecycle(new Promise());
    expect(result).toEqual(something);
  });
});
```

PWKad:
Alright I added a few comments, thanks again a ton for your help on this.

src/dialog-utilities.js
mbroadst: I may be mistaken, but this should still be private to the module. I'm not exporting it in `index.js`. Agreed though. As I mentioned above though, this method was already not fully tested in that the `compositionEngine` branch was never triggered

src/dialog-utilities.js
mbroadst: unless I'm missing something here, I haven't changed the logic of `invokeLifecycle`, I just reordered it to bail early (reduce indentation). The main difference here is the code used to be:
```
if (typeof instance[name] === 'function') {
  /* all of the actual logic with return statements */
}

return Promise.resolve(true); // the default case
```
and was changed to:
```
if (typeof instance[name] !== 'function') {
  return Promise.resolve(true); // bail out early reduce indentation by 1
}

/* all the actual logic */
```

PWKad: Typically we try to add test coverage as things are refactored, especially something like this where we move a utility in to a separate file from where it's used.  It helps to document the code through tests also so the next person who wants to submit a PR can get a better understanding of what each piece does and how it works together.

If you can document the behavior through failing specs at least I can help to get them passing.

EisenbergEffect:
@PWKad Is this ready to be merged? I'd like to update the modal dialog system to use the new CompositionTransaction capabilities but didn't want to do it until after we had all the PRs merged in from community members.

PWKad:
@EisenbergEffect No when we add new functionality or change existing functionality, even if it's just refactoring, we need to add tests to cover the changes.  I've offered to help @mbroadst with this so hopefully we can get this wrapped up soon.  I'm not aware of the CompositionTransaction capabilities so if you can link to that it may help to figure out what is needed from that regard while we are wrapping up any other changes.

EisenbergEffect:
The CompositionTransaction will enable us to delay showing the dialog until all its child compositions are completed. I can add it after this work is done.

mbroadst:
@PWKad @EisenbergEffect sorry I was away on vacation, I'll take a look at adding tests for these changes.

The reason I put this together was because I was working on a toastr implementation for aurelia, where I reuse some of this code (`invokeLifecycle` and `_getViewModel` in particular), does it make sense to think about how to move these type of utility methods somewhere "higher in the chain"? I know for a fact that `aurelia-notification` copies these methods as well, so that sort of completes our "rule of three" for refactoring :smile:

PWKad:
@mbroadst I think that definitely seems reasonable.  I'm not sure the best plan for refactoring to make these re-usable yet though, any ideas?

EisenbergEffect:
@mbroadst Note that you will need to rebase this on master since there have been some significant changes since this PR was made.

EisenbergEffect:
I'm going to close this for now since there was some major work done here that conflicts with this PR. If you feel there is more that needs to be done, please take a look at master and submit a new PR based on that. Thanks!

PR #120, opened 2016-03-18
refactor(dialog): Allow multiple renderers

### LINKED ISSUE #119: Refactor dialog-controller and dialog-renderer for better separation of concerns
I think that the `aurelia-dialog` plugin should be able to support multiple renderers. This would mean that the dialog plugin would function as a controller-style backend for any modal dialog implementation.

The problem right now is that there is some functionality wrapped in the `dialog-renderer` module, specifically in `createDialogHost` that I believe should be factored into `dialog-controller` or another module. If we can create the viewSlot outside of the renderer and pass it in, then we can easily swap in different modal implementations. Additionally, if we create a base interface for the renderer component, we can remove specific calls in the controller (like destroyDialogHost) and replace them with interface equivalents.

Even if there are other solutions down the line, this may lower the barrier to entry for people who want to use Aurelia viewmodels with different modal plugins; they can create their own renderer (or choose one that already exists) and get working quickly. Any thoughts?

jedd-ahyoung:
The dialog plugin was previously coupled to a single renderer. We refactor the dialog-controller and the render interface to provide a render abstraction against which multiple renderers can be defined. We alter the configuration builder to allow a user to choose a render implementation.

Fixes #119.


jedd-ahyoung:
I'd like some feedback on this PR and a bit of peer review. Right now, I've built this because we need it for my work project (I'll have to build a custom renderer to work with a different modal plugin). Things that I'm aware that this needs:

1. Code commenting, especially for the abstract renderer class.
2. Better naming conventions for files and variable names.
3. Possibly a better interface abstraction? Perhaps it might be prudent for the render interface to match the HTML5 dialog spec?
4. The configuration builder is a little rough.
5. Tests for the configuration builder. (Yeah, yeah, bad Jedd, I know.)
6. Typescript definitions.

Any thoughts are appreciated.

jedd-ahyoung:
@EisenbergEffect @PWKad @Aaike I think we're going to pull this into our work project to integrate with different modal solutions (perhaps the Bootstrap Modal plugin). If any of you have a chance, I'd love to know whether this is a direction you'd like to take with the dialog module.

EisenbergEffect:
Let me review now.

src/dialog-configuration.js
```diff
@@ -0,0 +1,38 @@
...
+      .useResource('ai-dialog-footer')
+      .useResource('attach-focus');
+  }
+
+  useResource(resourceName) {
+    this.aurelia.globalResources(resources[resourceName]);
+    return this;
+  }
+
+  useRenderer(renderer, settings) {
```
EisenbergEffect: Is there a reason for the settings param here?

EisenbergEffect:
This looks pretty nice. Did you test this in your project and with the sample app?

src/dialog-configuration.js
jedd-ahyoung: Whoops. There's supposed to be an `Object.assign()` statement in that block to merge the settings with the `globalSettings` object so that you can organize settings with the renderer itself. (They are all global settings, though.)

EisenbergEffect: Ok, if you can fix that and amend the PR...and confirm you've done some testing on this, I can merge it today. Release would happen tomorrow most likely.

jedd-ahyoung:
Yes. Tests are passing and the sample app is working. I've pulled this into my project at work and I haven't seen any discrepancies. I specifically made sure that using the dialog without configuration should yield the same results. Configuring the dialog plugin is slightly different, though - instead of this:
```
  aurelia.use
    .plugin('aurelia-dialog', (config) => {
      config.lock = true;
      config.centerHorizontalOnly = false;
      config.startingZIndex = 5;
    }); 
```

...we have to do this:
```
  aurelia.use
    .plugin('aurelia-dialog', (config) => {
      config.useDefaults();
      config.settings.lock = true;
      config.settings.centerHorizontalOnly = false;
      config.settings.startingZIndex = 5;
    }); 
```

Additionally, the plugin no longer exports the dialog custom elements unless you specifically call `useDefaults` (if you are providing a configuration function). I'm not sure if this might be potentially confusing in terms of migration.

jedd-ahyoung:
@EisenbergEffect Also, note that this is a proof of concept - it still needs to be cleaned up a bit. I just wanted to know if we were thinking of going in a different direction. I'd like to keep our project in close parity with the Aurelia core libraries, if possible.

EisenbergEffect:
Gotcha. I think this all looks good. If you can polish it up, we can merge it in. Just ping me when you think it's good to go.

jedd-ahyoung:
Sure, will do.

jedd-ahyoung:
@EisenbergEffect This should be ready.

PWKad:
So should use defaults fall back to a single renderer and configuration allows multiple or vice versa?  I'm good either way just curious

jedd-ahyoung:
@PWKad Ideally, the defaults should allow people who are currently using the plugin to continue using it without having to change any code. As I see it, the plugin will always use a single renderer (as it's a singleton) but the render implementation can be swapped out application-wide. Essentially, the defaults fall back to the renderer we provide now. In the future, we may be able to provide a renderer for, say, the HTML5 dialog element, and a dev can configure that and use it. Third-party renderers (such as the bootstrap dialog renderer I created for our work application) can also be plugged in.

I'd imagined a single renderer app-wide, but I believe it should be possible to allow multiple renderers via configuration. I'm not sure what the use case would be for that, necessarily, but I suppose that if we wanted we could pass an argument to the open method in the dialogService to choose a different renderer per dialog instance. Actually, perhaps a use case for this could be a browser modal plugin versus a mobile-specific modal plugin - for, say, phonegap applications. What do you think? Do you think there might be a use case for this?

PWKad:
Yes I agree I thought you meant the default was to not use a single so you are right.  Thanks!

sebastien-roch:
Hi,
I'm looking for some documentation about custom renderers, is there any? My goal is to make my modal draggable and resizable. Thanks

Issue #128, opened 2016-03-27
Close button does not appear to respect padding anymore


PWKad:
![image](https://cloud.githubusercontent.com/assets/4238393/14065455/a5119160-f3ef-11e5-811c-d799766cc35a.png)

jedd-ahyoung:
This is just a float issue. It's working as intended - it looks when when there is text in the header, but without it the header collapses because the button has been removed from the document flow. The real solution here is to remove the float.

Issue #103, opened 2016-03-04
Dialog will not scroll when dialog height is larger then containing page


airboss001:
If the page is shorter in height than the dialog that needs to be displayed the dialog gets cut off and you cannot scroll the dialog into view.

I am not a CSS person at all, but was able to hack at it some and get it working for my purposes by changing the following 3 css classes in dialog.css:

ai-dialog-overlay {
  bottom: 0;
  left: 0;
  position: absolute; //changed from fixed
  top: 0;
  right: 0;
  opacity: 0;
  pointer-events: none;
}

ai-dialog-container {
  display: block;
  position: absolute; //changed from fixed
  transition: opacity .2s linear;
  top: 0;
  right: 0;
  bottom: 0;
  left: 0;
  opacity: 0;
  // removed - overflow-x: hidden;
  // removed -  overflow-y: auto;
  -webkit-overflow-scrolling: touch;
}

.ai-dialog-open {
  overflow: scroll; //changed from hidden
}

PR #141, opened 2016-04-22
Renderer bug fixes - scrolling and close behavior

### LINKED ISSUE #125: Dialog doesn't close on click out
Hi, according to the docs (and the source), if `setting.lock = false` and the user clicks outside of the dialog, it will close it and resolve it with cancel. But, when I click outside - nothing happens.

The event handler, for closing the dialog, is set on the overlay layer (https://github.com/aurelia/dialog/blob/master/src/renderers/dialog-renderer.js#L76) and is never called, because `ai-dialog-overlay` has css style `pointer-events:none`, which prevents all clicks on this element. 

I found this issue #74, which describes the same problem about closing the dialog, but the problem were that the container was on top of the overlay and it was fixed with disabling pointer events on the `ai-dialog-container` to enable click events on `ai-dialog-overlay` #79

When I remove `pointer-events:none`; from `ai-dialog-overlay` and set it on `ai-dialog-container` the event handler is called and the dialog is closed correctly.

jedd-ahyoung:
Fixes #125, #103, and #128.


EisenbergEffect:
Awesome. @PWKad Can you review this. I like to always have a second team member review if possible. @jedd-ahyoung If Patrick approves, then go ahead and merge. Ping me after the merge. Keep me posted on all the updates so I can do a new release when things are ready.

jedd-ahyoung:
So I had to hack something together - there's a CSS way to get the overlay click working, but it doesn't work if you want to be able to scroll the modal window. As such, I defaulted to doing an "event.stopPropagation" on the ai-dialog element. However, I just realized - I don't know if this will cause problems with event.delegate functions in Aurelia, as they might need to bubble back up to the root element....

EisenbergEffect:
Yes, that would cause an issue. There might be a way around it though using
DOM boundaries. It's something we added for shadow DOM that might work here
too.

jedd-ahyoung:
After having looked into this problem, I think that the DOM boundary solution will work - but due to how the renderer and the service interact (and how the view is structured), I think that this solution will require a refactor of the default renderer component. This needs to happen anyway as the thing is pretty buggy as-is. (I believe that using transient renderers instead of a singleton will simplify everything and remove a lot of bugs.)

I can do this but I don't have immediate time. How quickly would you like a fix for this? If necessary, we can hack a patch together for a quick release and do a full refactor later. Let me know.

EisenbergEffect:
Give me an ETA and we can make a decision based on that. I don't see any issues with making the render transient. That seems to make sense, esp. if you want to be able to specify a per instance renderer in the future.

jedd-ahyoung:
@EisenbergEffect I authored https://github.com/aurelia/dialog/pull/143 on Saturday while en route back to Tallahassee. If that's accepted, I can rebase this change onto master and find a solution for the click issue.

PWKad:
This lgtm

EisenbergEffect:
@jedd-ahyoung Where are we at with this?

jedd-ahyoung:
@EisenbergEffect I've just forced pushed this branch with updates. I solved the overlay issue by monkey-patching the event and checking for the property duing the bubble phase. Tests are passing and this can be merged after it's rebased onto the other refactor PR.

PR #143, opened 2016-04-24
refactor(dialog): refactor renderer and service

### LINKED ISSUE #67: Using multiple modals
I might be doing something completely wrong, but I'm having issues with using aurelia-dialog multiple times throughout my application.

### Behavior
What's happening is:

1. I go to page foo, trigger opening a modal: works fine
2. I go to page bar, trigger opening a different modal: It fails with an error (described below)
3. I go back to page foo, trigger opening the modal again, and I get the content of page bar in stead.
4. This behavior now stays the same. The wrong content on page foo, and an error on page bar.

### Technically
I get the following error in the console:

> Unhandled promise rejection Error: No Aurelia APIs are defined for the referenced element.

The code I'm using is simply:

**overview.js:**
```javascript
this.dialogService.open({viewModel: Create}).then(response => {
  if (!response.wasCancelled) {
    this.companies.push(response.output);
  }
});
```

**create-modal.js:**
```javascript
import CompanyEntity from 'entity/company';
import {inject} from 'aurelia-framework';
import {Notification}  from 'service/notification';
import {Router} from 'aurelia-router';
import {DialogController} from 'aurelia-dialog';

@inject(CompanyEntity, Notification, Router, DialogController)
export class Create {

  requestInFlight = false;

  constructor (CompanyEntity, Notification, Router, DialogController) {
    this.notification                             = Notification;
    this.entity                                   = CompanyEntity;
    this.router                                   = Router;
    this.controller                               = DialogController;
    this.controller.settings.lock                 = false;
    this.controller.settings.centerHorizontalOnly = true;
  }

  create () {
    this.requestInFlight = true;

    this.entity.save()
      .then(result => {
        this.requestInFlight = false;
        this.notification.success('Location created successfully');
        this.controller.ok(result);
      })
      .catch(error => {
        console.error(error);
      });
  }
}
```

I'm using the default controller, and as far as I can tell I'm not doing anything _weird_.

### LINKED ISSUE #70: ai-dialog-open class removed with multiple dialogs
If you open more than one dialog, then closing just one of them will remove the `ai-dialog-open` css class from body.

The class should not be removed, if there's still a visible dialog.

jedd-ahyoung:
Change dialog renderer to use PAL. Alter DOM ordering of dialog elements and remove z-index ordering. Alter default renderer to be transient

Fixes #67. Fixes #70.

### DIFF
--- a/src/renderers/dialog-renderer.js
+++ b/src/renderers/dialog-renderer.js
@@ -1,69 +1,79 @@
 import {dialogOptions} from '../dialog-options';
+import {DOM} from 'aurelia-pal';
+import {Container, transient} from 'aurelia-dependency-injection';
 
-let currentZIndex = 1000;
+let containerTagName = 'ai-dialog-container';
+let overlayTagName = 'ai-dialog-overlay';
+let transitionEvent = function() {
+  let transition = null;
 
-let transitionEvent = (function() {
-  let t;
-  let el = document.createElement('fakeelement');
+  return transition || function() {
+    let t;
+    let el = DOM.createElement('fakeelement');
+    let transitions = {
+      'transition': 'transitionend',
+      'OTransition': 'oTransitionEnd',
+      'MozTransition': 'transitionend',
+      'WebkitTransition': 'webkitTransitionEnd'
+    };
+    for (t in transitions) {
+      if (el.style[t] !== undefined) {
+        transition = transitions[t];
+        return transition;
+      }
+    }
+  };
+};
 
-  let transitions = {
-    'transition': 'transitionend',
-    'OTransition': 'oTransitionEnd',
-    'MozTransition': 'transitionend',
-    'WebkitTransition': 'webkitTransitionEnd'
+@transient()
+export class DialogRenderer {
+  dialogControllers = [];
+  escapeKeyEvent = (e) => {
+    if (e.keyCode === 27) {
+      let top = this.dialogControllers[this.dialogControllers.length - 1];
+      if (top && top.settings.lock !== true) {
+        top.cancel();
+      }
+    }
   };
 
-  for (t in transitions) {
-    if (el.style[t] !== undefined) {
-      return transitions[t];
-    }
-  }
-})();
+  static inject = [Container];
 
-function getNextZIndex() {
-  return ++currentZIndex;
-}
-
-function centerDialog(modalContainer) {
-  const child = modalContainer.children[0];
-  const vh = Math.max(document.documentElement.clientHeight, window.innerHeight || 0);
-
-  child.style.marginTop = Math.max((vh - child.offsetHeight) / 2, 30) + 'px';
-}
-
-export class DialogRenderer {
-  defaultSettings = dialogOptions;
   constructor() {
-    currentZIndex = dialogOptions.startingZIndex;
-    this.dialogControllers = [];
-    this.containerTagName = 'ai-dialog-container';
-    document.addEventListener('keyup', e => {
-      if (e.keyCode === 27) {
-        let top = this.dialogControllers[this.dialogControllers.length - 1];
-        if (top && top.settings.lock !== true) {
-          top.cancel();
-        }
-      }
-    });
+    this.defaultSettings = dialogOptions;
+    this.dialogController = null;
   }
 
   getDialogContainer() {
-    return document.createElement('ai-dialog-container');
+    return DOM.createElement(containerTagName);
   }
 
-  createDialogHost(dialogController: DialogController) {
+  showDialog(dialogController: DialogController) {
+    if (!dialogController.showDialog) {
+      return this._createDialogHost(dialogController).then(() => {
+        return dialogController.showDialog();
+      });
+    }
+    return dialogController.showDialog();
+  }
+
+  hideDialog(dialogController: DialogController) {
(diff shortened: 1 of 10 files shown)


PWKad:
Was this fixing any issues?  Curious why we need to remove the z-index.  Also if we want to merge this in as-is would need to fix Readme where it references setting up zindex right?

test/unit/dialog-service.spec.js
```diff
@@ -37,7 +48,7 @@ describe('the Dialog Service', () => {
     dialogService.open(settings);
   });
 
-  it('calls position if specified', (done) => {
+  xit('calls position if specified', (done) => {
```
PWKad: Is this test no longer needed?

jedd-ahyoung:
Instead of relying on z-indexing, I used DOM ordering to make modal stacking work. The z-index incrementing was a little off, as well - closing all dialogs would never reset the current z-index to its initial starting value. I haven't removed the z-index entirely - it just no longer increments for each layer, instead using the starting z-index value for all dialog overlays and dialog containers.

This does cause an issue with the top-layer dialog not being the first child of the body element, causing potential tabbing issues - but as that's a separate issue that needs to be addressed (restricting focus to the current modal window, and setting aria-hidden on all other modals) I thought that this would be acceptable.

This happens to fix #70. This does not address #67 - an oversight on my part. I will amend the commit message.

PWKad:
Cool thanks, sounds good.

test/unit/dialog-service.spec.js
jedd-ahyoung: That specific test is somewhat problematic, as it spans the dialog-service, the dialog-controller, and the renderer itself. I intended to move it to the renderer but had troubles testing due to the separation of concerns. I left the test in place because it needs to be fixed. If this needs to be addressed for this PR, no problem - let me know.

src/renderers/dialog-renderer.js
```diff
@@ -1,7 +1,9 @@
...
+import {DOM} from 'aurelia-pal';
+import {Container, transient} from 'aurelia-dependency-injection';
 
-let currentZIndex = 1000;
-
+let containerTagName = 'ai-dialog-container';
+let overlayTagName = 'ai-dialog-overlay';
 let transitionEvent = (function() {
   let t;
   let el = document.createElement('fakeelement');
```
Gheoan: Shouldn't `DOM.createElement` be used here?

jedd-ahyoung: Whoops! Yes. Good catch.

EisenbergEffect: Yes. @jedd-ahyoung Please remove any document or window references and use the PAL instead.

EisenbergEffect:
How does this relate to the other dialog PR?

jedd-ahyoung:
@EisenbergEffect Due to the DOM boundary solution being a non-starter, it's no longer related directly.

EisenbergEffect:
Ok, so I can merge this independently? Should I merge the other as well? or wait on one or both...

jedd-ahyoung:
I'll correct the stuff in the code review first.

2016-04-26 12:00 GMT-04:00 Rob Eisenberg <notifications@github.com>:

EisenbergEffect:
Status?

jedd-ahyoung:
@EisenbergEffect This is complete and can be merged.

sensedeep:
You don't seem to be able to inject DialogController anymore. When injected, it has default/undefined args. The constructor references renderer:

```
      this.settings = Object.assign({}, this._renderer.defaultSettings, settings);
```

this then throws if renderer is undefined.

EisenbergEffect:
@jedd-ahyoung Please take a look into this.

sensedeep:
Does DialogController need to be manually invoked now rather than injected?  If so, the examples need updating in the repository. 

The constructor now requires 4 parameters:

    function DialogController(renderer, settings, resolve, reject)

Any example of using this available?

sensedeep:
I'm struggling to see how the current implementation can work for many use cases. DialogController is injected in quite a few places in aurelia-dialog. But when invoked as a new Type() from dep injection, it gets undefined exception on missing renderer.  How can this possibly work. I must be missing something fundamental in understanding how to use the new version of dialog.

EisenbergEffect:
Potentially you have some forks in your libraries? Can you check? Maybe it's a result of that.

sensedeep:
I've checked for forks and have upgraded to the very latest skeleton. I've been reading the code directly on the repository. Latest commit. This is all local to dialog and not involving other repos.

The issue is that DialogController is injected both by dialog and by user code. But the constructor requires 4 parameters that are not provided by DI.  Is that right?  Are you not meant to inject DialogController anymore?  How can dialog work then as it injects it without providing these parameters.

The prior version had parameters, but it worked when they were undefined because they were not directly sub-referenced in the constructor. This code did not exist:

```
this.settings = Object.assign({}, this._renderer.defaultSettings, settings);
```

previously it just assigned the renderer to this.renderer and if it was undefined ... no problem.

EisenbergEffect:
We'll look into this shortly. @jedd-ahyoung Please take a look. There may have been a regression.

sensedeep:
Thanks .. really appreciate the responsive attention you provide to all the issues. Gives us good confidence using Aurelia for a major app for production.

Issue #137, opened 2016-04-14
Fast code close crash


rockResolve:
When I open and close a dialog quickly using code, the close can cause a number of different errors depending on the timing. At the worst "hideDialog is undefined", when the controller is not fully built on open, but a close is attempted.

**Use case:** Progress or Activity dialogs that may be closed by code rather than by a user clicking a button. Sometimes they may be closed very quickly after they have been opened.

Both the Open & Close operations generate promises to do the work, and both also use the CSS TransitionEnd event to trigger further processing. These factors cause two problems:

1. The Open CSS TransitionEnd event is not always fired. Normally this is not noticed as no further Open processing occurs afterward. However this appears to also cause the Close TransitionEnd event to never fire either. As the TransitionEnd is in the middle of close processing, it in turn causes the dialog & overlay to be hidden, but not destroyed. The symptom is that the invisible overlay consumes all mouse clicks.

2. The current code does not publish the Open Dialog promise. Currently the open returns a promise resolved with the state of any later closure. However very quick closes must wait at least for the dialog to fully open, otherwise the partially open dialog causes errors when the close is processed.

PR to follow

jedd-ahyoung:
@rockResolve I've addressed this issue with a commit that allows skipping CSS transitions. However, I think we've also addressed the base issue by simplifying the renderer (no more method undefined errors). Within the next few days or so, would you be able to confirm that this is working for you (without ignoring the CSS transitions)? I'd like to see if we're supporting your use case with these fixes.

EisenbergEffect:
@jedd-ahyoung Is this resolved now?

rockResolve:
Sounds great in principal. Hope to test soon.

davismj:
@rockResolve did you have a chance to test? is the issue resolved?

rockResolve:
So finally had a chance to test this.

I found the root cause of much of this was the app router running removeAll on the viewSlot. This removed my previously created "Fetching data" dialog element, so the TransitionEnd was not fired, and the open (or openAndYeildController) promise was never resolved.

So ignoreTransitions was the perfect fix.

I also use the new dialogService.openAndYieldController which simplified my code.

Many thanks.

PWKad:
Just noticed this was actually closed.

Issue #96, opened 2016-02-18
Dialog not closed when browser back button is used


henning-krause:
Started with the skeleton-navigation app and added aurelia-dialog.

Added this import to the top of the welcome.js file:
```js
import {DialogService} from 'aurelia-dialog';
import {Prompt} from 'aurelia-dialog/./examples/prompt';
```

And this submit handler, taken from the Aurelia-Dialog documentation page:

```js
  submit() {
      this.dialogService.open({ viewModel: Prompt, model: 'Good or Bad?'}).then(response => {
      if (!response.wasCancelled) {
        console.log('good');
      } else {
        console.log('bad');
      }
      console.log(response.output);
    });
```

Repro:
1. Open the web page
2. Click on Github Users
3. Click on Welcome
4. Click Submit. The dialog is opened.
5. Click on the Back button of the browser.

Effect:
The "Github User" page is displayes again, but the dialog is still open.

Any ideas on how to solve this?

PWKad:
Good question, let me take a look and think about it.  Thanks for the repro steps!

piwnik:
While this issue isn't solved, is there any way to close dialog from parent? Or check if any dialog is shown (without tracking this by myself)? Then we probably can on parent canDeactivate check is the dialog shown, if yes - return false, but close dialog, and on second 'back' return true.

piwnik:
@henning-krause I solve this issue by adding ['autorize' router pipeline step](http://aurelia.io/docs.html#/aurelia/framework/1.0.0-beta.1.1.4/doc/article/securing-your-app).

```javascript
run(routingContext: NavigationInstruction, next) {

		// HACK closing dialog before navigate away from view
		var controller = this.dialogService.renderer.dialogControllers[0];

		if (controller) {
			controller.cancel();
			return next.cancel(new Redirect(routingContext.router.history.previousLocation, {trigger:false, replace: false}));
		}

		// HACKend
	
		return next();
}
````

Where if dialog is open - close it and stay on same page.

If you need only close dialog and continue navigation, just remove redirection line (which recovers history step)
```javascript
return next.cancel(new Redirect(routingContext.router.history.previousLocation, {trigger:false, replace: false}));

```

It's dirty hack, but I hope it solve your issue until real legal solution will be implemented by Aurelia team

PWKad:
[...]

PR #144, opened 2016-04-26
Fix show & close for fast code close of dialog

rockResolve:
For progress dialogs: allow fast code close. Fix for #137

Ensure show & close always complete
add options.showDialogTimeout
return model.showDialogPromise

Includes documentation changes and sample demos.
Sample demo includes faulty usage button, to show the problem if showDialogPromise is not used.

### DIFF
--- a/sample/src/progress.html
+++ b/sample/src/progress.html
@@ -1,0 +1,17 @@
+<template>
+  <require from='./progress.css'></require>
+  <ai-dialog>
+    <ai-dialog-body>
+      <div class='blockPageDialog'>
+        <div class='progress-message'>${waitMessage}</div>
+        <!-- <i class="fa fa-spinner fa-spin"></i>  -->
+        <div class="progress-bar progress">
+          <div css="width: ${completedPercent}%"></div>
+        </div>
+      </div>
+    </ai-dialog-body>
+    <ai-dialog-footer>
+      <button click.trigger="controller.cancel()">Cancel</button>
+    </ai-dialog-footer>
+  </ai-dialog>
+</template>
--- a/README.md
+++ b/README.md
@@ -111,16 +111,16 @@ jspm install aurelia-dialog
   ```
 3. Create (if you haven't already) a file `main.js` in your `src` folder with following content:
 
-```javascript
-  export function configure(aurelia) {
-    aurelia.use
-      .standardConfiguration()
-      .developmentLogging()
-      .plugin('aurelia-dialog');
+	```javascript
+	  export function configure(aurelia) {
+	    aurelia.use
+	      .standardConfiguration()
+	      .developmentLogging()
+	      .plugin('aurelia-dialog');
 
-    aurelia.start().then(a => a.setRoot());
-  }
-```
+	    aurelia.start().then(a => a.setRoot());
+	  }
+	```
 
 ## Using the plugin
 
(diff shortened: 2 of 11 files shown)


EisenbergEffect:
@jedd-ahyoung Can you review please?

sample/src/progress.html
```diff
@@ -0,0 +1,17 @@
+<template>
+  <require from='./progress.css'></require>
+    <ai-dialog>
+        <ai-dialog-body>
+            <div class='blockPageDialog'>
```
PWKad: Can you fix the tabs so that they are all two spaces?  (README.md also has mixed spacing)

jedd-ahyoung:
@rockResolve Thanks for the PR! I'll review it during the week. In the meantime, can you amend the commit to follow the [Aurelia commit guidelines](https://github.com/DurandalProject/about/blob/master/CONTRIBUTING.md#commit)?

rockResolve:
@PWKad @jedd-ahyoung 
I've amended the commit to fix your catches as requested.

jedd-ahyoung:
@EisenbergEffect  I've taken a look at this and it looks pretty good. @rockResolve I do have a question, though - explained here (https://github.com/aurelia/dialog/issues/137), you mentioned that closing the dialog before the open is finished causes issues. The solution in the PR fixes this, but requires client code to change (by adding the transition promise). I'm wondering if we can remove that necessity by simply storing the promise that the `showDialog` method returns - we can have the `hideDialog` method in the renderer only fire after that promise has resolved. (Something like - `return showDialogPromise.then(() => { return new Promise(...) });`) This would avoid having to change the interface to expose the transition. Do you think this would work for your use case while fixing the general issue?

Let me know what you think.

rockResolve:
@jedd-ahyoung That would improve things generally. 

For me I have multiple sequential progress dialogs (i.e. get generic startup data, then first view's data). So when opening a possibly open progress dialog I need to know the dialog's state, e.g is it opening. I am using the showDialogPromise for this. 

With your suggestion I could just go ahead and close anytime, using the dailogController.close returned promise to tell if the dialog is closed yet (or to wait for it to close before opening again). So your suggestion would work for me.

I cant think of a scenario where anyone would want to know for sure the dialog is fully open except to ensure it closes. Changing messages while a dialog would be handled by binding. Even timed dialogs (e.g. close after 2 minutes) wouldnt care if the timer was before or after the 200 ms transition.

Sounds like a good option as it removes the promise from the interface.

EisenbergEffect:
In light of the conversation here, I'm going to close this PR. Can someone submit a new PR with the proposed change. @jedd-ahyoung is that something you have time to do?

jedd-ahyoung:
@EisenbergEffect Sure. @rockResolve has actually done all of the legwork here. It'd just be a matter of removing the promise from `_settings.model` and storing it in the renderer.

EisenbergEffect:
Ok, sounds good. When you get a chance, just submit the new PR and we can merge. Are the other dialog PRs good to go? Should I get them merged first?

jedd-ahyoung:
No. The refactor has a small fix that I need to do, and I need to try to
fix the test coverage. The renderer changes for scrolling, etc, are still
suffering from the click issue that we'd discussed before.

2016-05-04 11:50 GMT-04:00 Rob Eisenberg <notifications@github.com>:

flieks:
Hi,

How can i close the dialog from within the viewmodel that launched the dialog (and received the promise)?

Greetings

rockResolve:
controller.close(), controller is made available during component activation.

For a summary have a look at the PR ReadMe.md, heading "Force close from code - Progress dialogs". 

For the detail see PR's sample folder. Look at app.js fastestProgress method. It uses closeDialog created by progress.js

flieks:
Sorry @rockResolve but what is PR ?
I only find this [readme.md](https://github.com/aurelia/dialog/blob/master/README.md)

rockResolve:
PR = Pull Request
See https://help.github.com/articles/using-pull-requests/

flieks:
Right i see, sorry.
But i don't need a progressbar, just want to close the dialog from the parent VM.

valichek:
I think this issue is closely related to closing dialog internally and having no access to dialog controller when opened. Currently it is not possible to get promise for dialog opening (transitions, animations, etc) outside. And I still think that my suggestion for API changes (https://github.com/aurelia/dialog/issues/96#issuecomment-219399087) is the only way to implement such scenarios but might be I'm wrong, so would be happy to get some working example.

valichek:
But the suggested behavior could be optional depending on dialogOptions passed

PR #162, opened 2016-06-03
doc(readme): cancel() requires a param.

HIRANO-Satoshi:


EisenbergEffect:
@jedd-ahyoung Should `cancel` require a param? I would think not. Can we see if there's a bug there or if we need to update some TS definitions to reflect that the param is optional?

EisenbergEffect:
Making the param optional in the d.ts.

Issue #157, opened 2016-05-25
It is no longer possible to inject the controller into a nested Dialogue


jsobell:
If a `DialogController` is injected in a nested Dialogue the load fails.
As such you can't access the `ok` or `cancel` methods from the nested control.

```
Unhandled rejection Error: Error invoking DialogController. Check the inner error for details.
------------------------------------------------
Inner Error:
Message: Cannot read property 'defaultSettings' of undefined
Inner Error Stack:
TypeError: Cannot read property 'defaultSettings' of undefined
    at new DialogController (webpack:///./~/aurelia-dialog/dist/commonjs/dialog-controller.js?:17:53)
    at Object.invoke (webpack:///./~/aurelia-dependency-injection/dist/commonjs/aurelia-dependency-injection.js?:322:12)
    at InvocationHandler.invoke (webpack:///./~/aurelia-dependency-injection/dist/commonjs/aurelia-dependency-injection.js?:299:166)
    at Container.invoke (webpack:///./~/aurelia-dependency-injection/dist/commonjs/aurelia-dependency-injection.js?:563:23)
    at StrategyResolver.get (webpack:///./~/aurelia-dependency-injection/dist/commonjs/aurelia-dependency-injection.js?:126:35)
    at Container.get (webpack:///./~/aurelia-dependency-injection/dist/commonjs/aurelia-dependency-injection.js?:494:39)
    at Object.invoke (webpack:///./~/aurelia-dependency-injection/dist/commonjs/aurelia-dependency-injection.js?:328:31)
    at InvocationHandler.invoke (webpack:///./~/aurelia-dependency-injection/dist/commonjs/aurelia-dependency-injection.js?:299:166)
    at Container.invoke (webpack:///./~/aurelia-dependency-injection/dist/commonjs/aurelia-dependency-injection.js?:563:23)
    at StrategyResolver.get (webpack:///./~/aurelia-dependency-injection/dist/commonjs/aurelia-dependency-injection.js?:126:35)
    at Container._get (webpack:///./~/aurelia-dependency-injection/dist/commonjs/aurelia-dependency-injection.js?:508:39)
    at Container.get (webpack:///./~/aurelia-dependency-injection/dist/commonjs/aurelia-dependency-injection.js?:497:26)
    at Object.invokeWithDynamicDependencies (webpack:///./~/aurelia-dependency-injection/dist/commonjs/aurelia-dependency-injection.js?:310:25)
[...]

PR #166, opened 2016-06-11
Fix renderer issues and DI issues with dialog-controller

jedd-ahyoung:


jedd-ahyoung:
@EisenbergEffect Do you have time to review this?

EisenbergEffect:
I can review shortly. I'm about to work on the dialog to update it to shadow dom v1 and fix some other issues.

src/dialog-controller.js
```diff
@@ -8,8 +8,10 @@ import {invokeLifecycle} from './lifecycle';
 export class DialogController {
   settings: any;
   constructor(renderer: DialogRenderer, settings: any, resolve: Function, reject: Function) {
+    let defaultSettings = renderer ? renderer.defaultSettings || {} : {};
```
EisenbergEffect: I don't understand this. Why/How could the DialogRenderer be null/undefined?

EisenbergEffect:
I have reviewed. I only had one question. @jedd-ahyoung Please respond asap so I can merge. I've got a lot of other work to do on this library and need to get it done asap...as in now because this library is completely broken with the latest releases due to the shadow dom update.

src/dialog-controller.js
jedd-ahyoung: This was related to https://github.com/aurelia/dialog/issues/157. I wasn't sure but it looked as though people were trying to manually inject the DialogController using a container that didn't have a DialogRenderer attached to it. I wasn't entirely sure about the use case, but this apparently worked before the renderer refactor (if I remember correctly, the controller didn't need a reference to the renderer, so the renderer wasn't a constructor argument). If this is a questionable change, I'm okay with removing it.

EisenbergEffect: I'm going to merge and then take a look at the whole codebase and see if there's something that can be improved here. Thanks!

PR #173, opened 2016-06-28
fix(dialog-configuration): Correct useRenderer typing

davismj:
This commit should fail today, as there is an issue with the babel-dts-generator (here https://github.com/YoloDev/babel-dts-generator/issues/54). However, the typing is accurate, and the current typing is in fact a mistake and throws compiler errors in TypeScript.

### DIFF
--- a/src/dialog-configuration.js
+++ b/src/dialog-configuration.js
@@ -64,7 +64,7 @@ export class DialogConfiguration {
    * @param settings Global settings for the renderer.
    * @return This instance.
    */
-  useRenderer(renderer: Renderer, settings?: Object): DialogConfiguration {
+  useRenderer(renderer: typeof Renderer, settings?: Object): DialogConfiguration {
     this.renderer = renderer;
     this.settings = Object.assign(this.settings, settings || {});
     return this;


jedd-ahyoung:
@davismj Does this essentially allow the type system to support subclasses? If there is currently an issue with the dts generator, it might be easier to make the Renderer a Typescript interface and abandon the "base class" idea altogether (as it doesn't provide a useful default anyway). What do you think?

davismj:
Using typeof Render _does_ add subclassing support, but this is just part of the deal. The bigger issue is that the argument there is not a `Renderer` instance. If you pass in a `new Renderer()`, it breaks. It might work if you pass in `container.get(Renderer)`, but this is more due to the flexibility of the DI container. In fact, the argument is a `typeof Renderer` since we want the DI container to register it in the `childContainer` found here: https://github.com/aurelia/dialog/blob/master/src/dialog-service.js#L31.

EisenbergEffect:
@jedd-ahyoung We are running out of time to work on these dialog issues. You have two weeks to wrap everything up basically.

jedd-ahyoung:
@davismj I'll admit - I'm a little lost here. 

> The bigger issue is that the argument there is not a Renderer instance. If you pass in a new Renderer(), it breaks.

Would this not be fixed by making Renderer an interface and having the default renderer "inherit" this interface? I'm not entlrely sure why Typescript isn't recognizing this (as I don't use TS). If this doesn't fix the problem, we're stuck waiting for babel upstream for a fix, unless we manually edit the dist files and do a point release.

davismj:
No. `useRenderer(renderer: Renderer, ...)` tells TypeScript that it is expecting an instance of a Renderer, where `useRenderer(renderer: typeof Renderer, ...)` tells TypeScript that it is expecting a Renderer _constructor_.

The `renderer` argument in fact is expected to be a `Renderer` constructor, as we pass this argument to the DI [here](https://github.com/aurelia/dialog/blob/master/src/dialog-service.js#L32).

```
dialogController = new DialogController(childContainer.get(Renderer), settings, resolve, reject);
```

aluanhaddad:
@davismj I'm not sure if babel-dts-generator supports these either, but you can also write the type as
```TypeScript
(new () => Renderer)
// or
{ new (): Renderer; }
// or with a type alias
type RendererConstructor = { 
  new (): Renderer; 
};
// or
type RendererConstructor = (new () => Renderer);
// or 
interface RenderStatic {
    new (): Renderer;
}
```
Just as with `typeof Renderer` none of these emit anything.
Not sure if this helps but these are all valid TypeScript annotations that say, renderer is a type with a construct signature that returns an instance of renderer. In any case, the only way to fulfill this contract from will be to provide either a `class` or `Function`. Maybe I don't understand the scenario.

aluanhaddad:
It turns out that the _babel-dts-generator_, which I guess is derived from Flow, does not support any of the alternatives I proposed. I was, however, able to get it working using a class based solution. I submitted this fix #178

davismj:
If we upgrade to the latest babel-dts-generator, including the fix here https://github.com/YoloDev/babel-dts-generator/issues/54, this should pass properly.

EisenbergEffect:
@davismj Is that all that is needed, merge this and update the dts generator?

davismj:
Yea, I'll work on this today.

EisenbergEffect:
Thanks @davismj Let me know.

PR #178, opened 2016-07-07
Workaround babel-dts-generator limits

aluanhaddad:
This creates a dummy class, `RendererBase` which simply extends `Renderer`. It provides a type declaration in a form that the _babel-dts-generator_ will accept. This fixes the issure in #173, Ideally, an interface would be used.


aluanhaddad:
Bah, I just realized this isn't going to solve the issue.

PR #184, opened 2016-07-15
fix(dialog-configuration): do not inject empty styling

RomkeVdMeulen:
Previously an empty `<style>` tag would be injected into the header if the
plugin was configured with no styling instructions.

### DIFF
--- a/src/dialog-configuration.js
+++ b/src/dialog-configuration.js
@@ -84,6 +84,8 @@ export class DialogConfiguration {
   _apply() {
     this.aurelia.singleton(Renderer, this.renderer);
     this.resources.forEach(resourceName => this.aurelia.globalResources(resources[resourceName]));
-    DOM.injectStyles(this.cssText);
+    if (this.cssText !== null && this.cssText.length > 0) {
+      DOM.injectStyles(this.cssText);
+    }
   }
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
Good catch. Thanks!

src/dialog-configuration.js
```diff
@@ -84,6 +84,8 @@ export class DialogConfiguration {
   _apply() {
     this.aurelia.singleton(Renderer, this.renderer);
     this.resources.forEach(resourceName => this.aurelia.globalResources(resources[resourceName]));
-    DOM.injectStyles(this.cssText);
+    if (this.cssText !== null && this.cssText.length > 0) {
```
CasiOo: Or simply `if (this.cssText)` to also catch undefined?

EisenbergEffect: Yes, that's probably a better implementation. Interested in submitting a PR?

CasiOo: After a good night's sleep, if @RomkeVdMeulen doesn't beat me to it first.

RomkeVdMeulen: Hadn't planned to :wink:

PR #187, opened 2016-07-16
fix(dialog-configuration): Skip injecting empty style

CasiOo:
Undefined style should not get injected.

Added check for undefined style.
Added tests for empty style.


EisenbergEffect:
Thanks!

Issue #190, opened 2016-07-19
"position" handler and no transitions issue


valichek:
When "position" handler is being set and no transitions for dialog defined in css, the promise returned from DialogRenderer.showDialog() resolves on dialog CLOSE! transition end.

EisenbergEffect:
@jedd-ahyoung Ping

EisenbergEffect:
@PWKad Are you able to look at this?

don-bluelinegrid:
Curious what the status is of this or https://github.com/aurelia/dialog/issues/208

We need to be able to do the following:
1. Open a dialog.
2. After user clicks 'OK', execute a fetch-client REST request.
3. On success of the request based on the fetch-client resolved Promise, then close the dialog; on failure of the request, show an error message in the dialog, which *should still be open*.

My understanding is that openAndYieldController() is designed for this use case, I.e., #3.  I've used code from the example for this, but the dialog closes immediately on calling ok() from the OK button click.  Is what I've described the same case as what's reported here, or something different?

Is there some other way to accomplish this?

Thanks,
Don

valichek:
@don-bluelinegrid to avoid this issue use `{ ignoreTransitions: true }` for options

valichek:
@don-bluelinegrid actually that depends on what your OK btn is doing, so smth different

don-bluelinegrid:
@valichek 

1. Where do we set ```{ ignoreTransitions: true }```?

2. 

> actually that depends on what your OK btn is doing, so smth different
My OK button is just calling ok().  How else can I use the OK button to communicate with teh DialogController to produce a result inside the Promise?

don-bluelinegrid:
@valichek 

I've tried ignoreTransitions, it made no difference with the code I am using.

Here is the code in my app:

```
    this.dialogService.openAndYieldController({
      viewModel: Model, 
      view: './communityModel.html', 
      model: {title: title, item: community, okText: this.i18n.tr('button.save')},
      ignoreTransitions: true 
    }).then((controller:any) => {
       controller.result.then((response) => {
    // .then(response => {
      if (!response.wasCancelled) {
        let community = response.output;
        if(community.commnunityId ) {
          // Update
        } else {
          // Create
          let comm = {
            communityId: community.communityId, 
            communityName: community.communityName, 
            communityDescription: community.communityDescription, 
            communityType: community.communityType,
            membershipType: 'DEFINED'
          };
          this.dataService.createCommunity(comm)
[...]

PR #191, opened 2016-07-19
Openc func

valichek:
Extend DialogService with openc func that returns promise for DialogController instance, the promise for result is stored in dialogController.result property


EisenbergEffect:
@PWKad @jedd-ahyoung What do you guys think of this api?

src/dialog-service.js
```diff
@@ -82,6 +53,68 @@ export class DialogService {
       return result;
     });
   }
+  
+  openc(settings?: Object): Promise<DialogController> {
```
PWKad: Can we give this a more descriptive name?

EisenbergEffect: Probably something like `openController()` yes?

PWKad:
I like it could use a more descriptive name but otherwise I think this solves many questions that have been raised.  @valichek Can you add some tests to cover the new functionality?  Some pseudo examples - 

> expect openc to return the correct instance of the controller
> expect calling close on returned controller to resolve the promise
> expect controller.cancel to throw an error if the dialog is already closed

src/dialog-service.js
PWKad: Not sure I was thinking that but it doesn't seem to properly describe what is being done, perhaps `openSynchronously`

EisenbergEffect: Yeah, I'm not sure. It's kind of tricky to name it.

valichek:
@PWKad yes, I will cover it with tests, the only problem I have, is #190 not fixed. Actually this issue have raised when I started the PR

PWKad:
@valichek understood I am guessing that is somehow related to this but reading that issue I'm not familiar with the changes that describes.  For my info, how is that blocking writing tests for this PR?

valichek:
@PWKad sorry, just translation issue) it's blocking for the PR

jedd-ahyoung:
This is a good addition, aside from the naming. I'd argue that this new API is actually better than the old one, as it's more flexible and can cover all of the use cases of the original "synchronous" API. Aside from naming and tests I think this should be good to go.

EisenbergEffect:
@jedd-ahyoung Is this ok to merge? We can rename the method after merge.

EisenbergEffect:
@jedd-ahyoung The release is tomorrow morning. Yes or no please.

Issue #180, opened 2016-07-13
Parents of nested Dialogs no longer close after child has closed


jsobell:
Something in the last update appears to have broken the closing of nested dialogs.
If a child dialog is displayed and closed, the controller of the parent no longer closes the parent.
It appears that renderer.hideDialog is called, but the parent doesn't close, so I assume it's something in the renderer failing to remove the dialog from the DOM.

EisenbergEffect:
@jedd-ahyoung We need to get on this quickly and put out a patch release. Looks like there are several regressions in that last update.

simonfox:
@EisenbergEffect I've opened the above PR which fixes this issue.

davismj:
I've reviewed the PR and even though circle ci reports a failed test, I've found that the test actually pass.

However, I'm not sure about this PR. Just from reading through the code, it seems that the `Renderer` was designed to be a singleton, not a transient, i.e. there was no mistake. Still, I see the value in making it a transient. 

Before we pull it in, I'd either like (a) @jedd-ahyoung to weigh in or (b) a sample gist / repo where I can reproduce the issue.

simonfox:
@davismj see the message with commit 0fcd535 by @jedd-ahyoung the renderer was refactored to be used as transient. 

Happy to also provide a gist if you still want it?

davismj:
Ah, on second look, I see what you're referring to. I'd still want @jedd-ahyoung to weigh in if possible, but it seems that this will likely be pulled in.

simonfox:
OK cool thanks. This definitely fixes the issue for us.

danhowitt:
@simonfox thanks for the fix. I've pulled this in to one of our projects to resolve the issue. Appears to work perfectly.

PR #192, opened 2016-07-20
fix(dialog-configuration): register renderer as transient

simonfox:
The renderer is now intended to be used per opened dialog (see 67e84f0). The configuration of the dialog plugin was not updated with that change and the registration of the renderer was still singleton. This PR updates the registration to transient.

closes aurelia/dialog#180

### DIFF
--- a/src/dialog-configuration.js
+++ b/src/dialog-configuration.js
@@ -82,7 +82,7 @@ export class DialogConfiguration {
   }
 
   _apply() {
-    this.aurelia.singleton(Renderer, this.renderer);
+    this.aurelia.transient(Renderer, this.renderer);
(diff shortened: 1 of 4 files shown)


EisenbergEffect:
@jedd-ahyoung Please review and merge if it's ok.

davismj:
I have confirmed that the tests pass, and other confirms that this fixes the issue. From reading through the code, there does seem to have been a bug and this does seem to resolve it, so I recommend merging it.

EisenbergEffect:
Thanks!

EisenbergEffect:
Got it. Thanks @simonfox and @davismj !

PR #194, opened 2016-07-23
fix(dialog-configuration): Fix useRenderer typing

davismj:


davismj:
Includes an update of the package.json to the latest babel-dts-generator, which is required for this commit to build. Replaces https://github.com/aurelia/dialog/pull/173.

doktordirk:
you lost the entries for jspm17 there. that.s on purposes?

davismj:
Good catch. Fixed.

PR #205, opened 2016-08-26
fix(dialog-service): .open() proper error propagation

StrahilKazlachev:
Correcting the promise chain in `.open()` to properly propagate errors.

### DIFF
--- a/src/dialog-service.js
+++ b/src/dialog-service.js
@@ -100,6 +100,9 @@ function _openDialog(service, childContainer, dialogController) {
           dialogController.view = controller.view;
 
           return dialogController.renderer.showDialog(dialogController);
+        }).catch(e => {
+          _removeController(service, dialogController);
+          return Promise.reject(e);
         });
       }
     });
--- a/src/dialog-service.js
+++ b/src/dialog-service.js
@@ -34,26 +34,27 @@ export class DialogService {
    * @return Promise A promise that settles when the dialog is closed.
    */
   open(settings?: Object): Promise<DialogResult> {
+    let childContainer = this.container.createChild();
     let dialogController;
-
     let promise = new Promise((resolve, reject) => {
-      let childContainer = this.container.createChild();
       dialogController = new DialogController(childContainer.get(Renderer), settings, resolve, reject);
-      childContainer.registerInstance(DialogController, dialogController);
-      return _openDialog(this, childContainer, dialogController);
     });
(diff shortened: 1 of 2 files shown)


PWKad:
@StrahilKazlachev awesome, thanks for this!  Can you give me a brief description of the error this PR fixes along with a quick code sample to reproduce?  Also please make sure if you haven't already to sign the CLA.  Thanks again!

StrahilKazlachev:
@PWKad  `.open()` was just returning the promise that resolves when the controller methods were called(`.ok()`, `.error()`, ...). The promise returned from `_openDialog()` was not chained and in case of errors while creating the VM('incorrect moduleId, errors thrown in the constructor or `.activate()`, ...) were not propagated to the caller of `.open()`. The dialog was never shown but the returned promise is never resolved. Not very brief, sorry. I can make a sample when I get back from work.

StrahilKazlachev:
While I was writing the above I thought that it would be better to `.catch()` in `_openDialog()` and calling `.error()` of the controller, thoughts?
```
function _openDialog(service, childContainer, dialogController) {
  ...
  return _getViewModel(instruction, service.compositionEngine)
    .then(returnedInstruction => {...})
    .catch(e => { dialogController.error(e); });
}
```

StrahilKazlachev:
@PWKad Here is a [sample](https://github.com/StrahilKazlachev/dialog/tree/sample/open/error-propagation). I've modified the dialog sample with 3 cases - incorrect moduleId, error thrown in the VM constructor and error thrown in the `.activate()` method. Check `app.js` what is expected to be logged in the console for every case. Also in the case of error thrown in the `.activate()` method you can see that no dialog is shown but the service `.hasActiveDialog` says `true`.

PWKad:
Seems reasonable to catch the errors.  Going over the sample now, I am trying to verify this doesn't break any currently working functionality.  As a note there don't appear to be any assertions in the unit tests.

PWKad:
I'm good with this @StrahilKazlachev thanks for the contribution.  @EisenbergEffect can you please confirm CLA?

EisenbergEffect:
@PWKad CLA signature is confirmed :)

PWKad:
@StrahilKazlachev let me know what you think about updating the test this way (I know it's brittle but at least we have an assertion in there) - https://github.com/aurelia/dialog/commit/79e39a76067f949a49b30e68bce6aa2570812d4a

Issue #189, opened 2016-07-18
Dialog isn't closed after canDecativate => false


piwnik:
If dialog VM canDeactivate function returns false, later on isn't possible close dialog at all.

I thing bug is here (dialog-controller.js, line 61):

```
 close(ok: boolean, output?: any): Promise<DialogResult> { 
       if (this._closePromise) return this._closePromise; 

       this._closePromise  = (....)
}
````

EisenbergEffect:
@jedd-ahyoung Ping.

StrahilKazlachev:
@PWKad This is merged.

PR #206, opened 2016-08-30
fix(dialog-controller): clear close promise when canDeactivate=>false

StrahilKazlachev:
fix for #189

### DIFF
--- a/src/dialog-controller.js
+++ b/src/dialog-controller.js
@@ -58,7 +58,9 @@ export class DialogController {
    * @returns Promise An empty promise object.
    */
   close(ok: boolean, output?: any): Promise<DialogResult> {
-    if (this._closePromise) return this._closePromise;
+    if (this._closePromise) { 
+      return this._closePromise;
+    }
 
     this._closePromise = invokeLifecycle(this.viewModel, 'canDeactivate').then(canDeactivate => {
       if (canDeactivate) {
@@ -73,7 +75,10 @@ export class DialogController {
           });
       }
 
-      return Promise.resolve();
+      this._closePromise = undefined;
+    }, e => {
+      this._closePromise = undefined;
+      return Promise.reject(e);
     });
 
     return this._closePromise;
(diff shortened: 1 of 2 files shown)


src/dialog-controller.js
```diff
@@ -58,7 +58,9 @@ export class DialogController {
    * @returns Promise An empty promise object.
    */
   close(ok: boolean, output?: any): Promise<DialogResult> {
-    if (this._closePromise) return this._closePromise;
+    if (this._closePromise) { 
+      return this._closePromise;
+    }
 
     this._closePromise = invokeLifecycle(this.viewModel, 'canDeactivate').then(canDeactivate => {
```
StrahilKazlachev: I'm not sure what should be done if the promise returned from `invokeLifecycle` is rejected. Should `_closePromise` still be cleared. Or call `.error()`?

PWKad: After reviewing this for a while I don't think it makes a difference so let's call `error` to be on the safe side.  Agreed?

StrahilKazlachev: @PWKad When I wrote it I wasn't sure if unresolved promises cause memory leaks and that was my reasoning for wanting to call `.error()` - checked, don't cause.
Also `.error` will call `invokeLifecycle(this.viewModel, 'deactivate')`. Is it OK for `deactivate` to be called when `canDeactivate` has failed? If it's OK, I'll test it with calling `.error`.

PWKad:
Gotcha I see what you mean now, if the promise throws I don't think calling `error` is proper like you mentioned.  Clearing it should be sufficient for now.  Long term might not hurt to have the state be a little bit easier to understand of why it's adding and clearing promises like that.

src/dialog-controller.js
```diff
@@ -73,7 +75,10 @@ export class DialogController {
           });
       }
 
-      return Promise.resolve();
```
PWKad: Was this not needed at all before?

StrahilKazlachev: I think so. At least I do not see any meaning.
 1.  we are in the success callback
 2. `canDeactivate` is `false`
 3. we return
  - no `return` statement
  - `return Promise.resolve()` 
 4. `_closePromise` will be a promise that is resolved without a result

EisenbergEffect:
@PWKad Is this ready to merge?

StrahilKazlachev:
@PWKad ping

PR #209, opened 2016-09-14
fix(dialog-renderer, dialog-service): fix settings merging

### LINKED ISSUE #181: config.settings.lock has no effect
After replacing all my custom headers with the Aurelia ones, I find I can't get rid of the 'X' to close the forms.
The value in config.settings.lock no longer has any effect, but the lock can be forced by passing it in the object to dialogServce.open().

StrahilKazlachev:
fix #181 - it turns out that the `if` binding in the `dialog-header` is messing with the `lock` setting. The binding gets to run before the renderer merges the default settings with the passed ones. So when `lock` is not passed:

1. a property for `lock` is installed on the `controller.settings` by the binding, and the value is `undefined`.
2. `rendered.showDialog` runs and merges the settings, the `undefined` overwrites the default value - not what we want, and the `controller.settings` are not updated.

Refactoring:
- moved the merging of the settings in the `dialog-service` so they can be merged as early as possible - there are other boolean settings that can get messed up if used in bindings.
- made `startingZIndex` not to be overwritten by settings passed to `open` and `openAndYieldController`, only when configuring the plugin - the render was using it from `defaultSettings` not the merged ones, it doesn't make much sense to change this per dialog basis.
- updated the tests accordingly - I hope so :laughing:


StrahilKazlachev:
@PWKad ping
The tests have failed at `jspm install`, how to re-trigger them?

PWKad:
Not 100% sure let me go check

PWKad:
Just needed to log in to rebuild it seems.  Done looks like it's passing again.

PWKad:
@StrahilKazlachev can you run `gulp lint` and fix the issues in `dialog-service.js` and any new ones?  It looks like at some point there were two created in master so if you want to include them in this PR that would be fine else I'll fix them (just trying to not to create a merge issue yet)

PWKad:
Really nice help btw on the PR's recently :)

StrahilKazlachev:
@PWKad Rebased and fixed the linting problems, didn't know there was a lint task.

PWKad:
Awesome thanks.  I'm going to add it to the CI process at least for dialog.  Thanks again for all your hard work.

Issue #211, opened 2016-09-23
Pressing escape closes all the dialogs


VMBindraban:
**I'm submitting a bug report**

* **Library Version:**
1.0.0-beta.3.0.0

* **JSPM**
JSPM 0.16.15

* **Browser:**
Chrome 52 / Firefox 48

* **Language:**
ESNext

**Current behavior:**
Pressing escape closes all the dialogs

**Expected/desired behavior:**
Pressing escape should only close the top dialog.
https://github.com/aurelia/dialog/blob/master/src/dialog-renderer.js#L35

EisenbergEffect:
@PWKad Can you take a look at this please?

PWKad:
Was hoping to get some tests or an explanation from Jedd on what has been done since the code is a bit of a mess but I'll do my best to figure out what's going on here

StrahilKazlachev:
@PWKad Actually I am working on this, but wanted to see how my other PR goes first.

PWKad:
Fixed in #212

Issue #207, opened 2016-09-04
ai-dialog-open class must be removed only when the last dialog closes.


ppn2:
**I'm submitting a bug report**

* **Library Version:**
1.0.1

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
4.4.6

* **NPM Version:**
3.10.6

* **JSPM OR Webpack AND Version**
requirejs 2.2.0, jspm 0.16.42

* **Browser:**
Chrome Version 52.0.2743.116 m
* **Language:**
TypeScript 1.8

**Current behavior:**
When a dialog opens, ai-dialog-open class is added to the body to hide the overflow. When dialog closes the ai-dialog-open is removed. However, if multiple dialogs are opened, the ai-dialog-open class should be removed only if the last dialog is opened. At this moment the class is removed as soon as first dialog is closed.

**Expected/desired behavior:**

if there are multiple dialogs, the ai-dialog-open class should be removed from the body tag, only if the last dialog is closed. Closing a dialog, while there are other dialogs underneath, should not remove the ai-dialog-open.

CasiOo:
I opened an issue about the same here https://github.com/aurelia/dialog/issues/70
I still find it being a bug.

ppn2:
hi, is there any progress on this?

PWKad:
Fix should be going out Tuesday

PR #212, opened 2016-09-26
fix(dialog-renderer): do not track controllers per renderer instance

### LINKED ISSUE #199: ESC key closes child and parent dialog
**I'm submitting a bug report**

* **Library Version:**
1.0.0-beta-3.0.0

* **Browser:**
Chrome, Firefox, Edge, IE 11

**Current behavior:**
I have a parent dialog and a child dialog. When the child dialog is on top of parent dialog, pressing the ESC key closes the child and parent dialog.

Pressing one ESC key should only close the child (topmost) dialog and not the parent dialog.

StrahilKazlachev:
fix #199, #207, #211 
From what I understand, currently the relationship renderer:controller should be 1:1, but the renderer is setup to track multiple controllers per instance. So when `ESC` is pressed every renderer closes its last controller - all controllers/dialogs get closed.

### DIFF
--- a/src/dialog-controller.js
+++ b/src/dialog-controller.js
@@ -58,7 +58,7 @@ export class DialogController {
    * @returns Promise An empty promise object.
    */
   close(ok: boolean, output?: any): Promise<DialogResult> {
-    if (this._closePromise) { 
+    if (this._closePromise) {
       return this._closePromise;
     }
 
--- a/src/dialog-renderer.js
+++ b/src/dialog-renderer.js
@@ -1,4 +1,3 @@
-import {dialogOptions} from './dialog-options';
 import {DOM} from 'aurelia-pal';
 import {transient} from 'aurelia-dependency-injection';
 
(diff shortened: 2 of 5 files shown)


EisenbergEffect:
@PWKad Can you review?

PWKad:
@StrahilKazlachev could you add a test to show multiple dialogs and the expected behavior passing after the PR?  I'm reviewing the open issues and if I understand correctly it seems that the issue is that the dialogs are actually all hidden, not removed, which seems to be because to be because dialogControllers on the instance is 1:1 so you should only ever have 1 instance of a dialogController in the dialogControllers property previously.  Now that those are moved out it almost seems like we should have a singleton to handle the state (is that over-engineering?)

If we can put together a set of tests showing that - 

1. In one view model, opens one dialog.  Invoking cancel on it should remove the dialog open class.
1. In one view model, opens two dialogs.  Invoking cancel on the second should not close the first, nor remove dialog open class.
1. In two view models, opens one dialog each.  Invoking cancel on the second should not close the first, nor remove dialog open class.

PWKad:
By the way if you need a hand on writing the tests or you want to pair up at some point on this let me know and we can discuss further (on gitter @ `pwkad`)

PWKad:
Looks great man thanks again!

PR #215, opened 2016-10-01
fix(dialog-service): fix remove controllers closed with .error

StrahilKazlachev:
- fix dialogs closed with `DialogController.error` should also be removed from `DialogService.controllers`
- fix controllers should be added to `DialogService.controllers` after the view model has activated(after `CompositionEngine.compose()` succeeds) - if not why is `DialogService.hasActiveDialog` computed from `DialogService.controllers.length`?
- refactor `DialogService.open()` to use `DialogService.openAndYieldController()` internally - nice for testing
- fix `reports no active dialog if no dialog is open` test
- add more tests

### DIFF
--- a/src/dialog-service.js
+++ b/src/dialog-service.js
@@ -35,19 +35,8 @@ export class DialogService {
    * @return Promise A promise that settles when the dialog is closed.
    */
   open(settings?: Object): Promise<DialogResult> {
-    let childContainer = this.container.createChild();
-    let dialogController;
-    let promise = new Promise((resolve, reject) => {
-      dialogController = new DialogController(childContainer.get(Renderer), _createSettings(settings), resolve, reject);
-    });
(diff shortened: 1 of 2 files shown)


StrahilKazlachev:
@PWKad One thing that is bugging me is that with the current `.d.ts` generation process we can not annotate the return type of `DialogService.openAndYieldController()` as `DialogController & { result: Promise<DialogResult>}`. What do you think of adding a `result: Promise<DialogResult>` field to the `DialogController`?

EisenbergEffect:
@PWKad Please review :)

EisenbergEffect:
@PWKad @jedd-ahyoung or @davismj Please review.

PWKad:
@StrahilKazlachev is the issue that the .d.ts doesn't respect union types or is it that the return type of `_openDialog` is incorrect or something else?

Issue #214, opened 2016-09-29
[ENHANCEMENT] Ability to close any open dialog upon route change.


don-bluelinegrid:
**I'm submitting a feature request**
- **Library Version:**
  1.0.0-beta.3.0.0

**Please tell us about your environment:**
- **Operating System:**
  OSX 10.5
- **Node Version:**
  4.4.7
- **NPM Version:**
  3.10.3
- **JSPM OR Webpack AND Version**
  webpack 2.1.0-beta.22
- **Browser:**
  all | Chrome XX | Firefox XX | Edge XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView
- **Language:**
  TypeScript X.X 

**Expected/desired behavior:**
There is a common issue that occurs when showing modal elements (elements attached to the DOM programmatically, outside the context of a view template) in a SPA.  If the underlying view is changed via navigation, the modal element won't be hidden.

As an example, let's say I have an Aurelia application with various routes registered, and I have visited these routes:
/#/home
/#/main
/#/manage
While on the /#/manage, view, I show a modal dialog using aurelia-dialog, and then I use the browser 'back' button (or the Router navigation) to return to the /#/main view.  In this case, I will see the /#/main view, with the dialog from the /#/manage view still showing.

It would be desirable to have the dialogs dismissed by default on a router view change, or at least have an option for this behavior.  Without this, some messy code has to be inserted to listen for route changes, check to see if there is a visible dialog, and then dismiss it.

Don

StrahilKazlachev:
I wouldn't say `messy`. The `dialog-service` has a `controllers: DialogController[]` property you can iterate and call `.cancel()` on every controller. Still there could be a utility method - `releaseAll()`, `closeAll()`, ... Would you format the issue according to the template to have it as a feature request?

don-bluelinegrid:
Regardless of the subjective view of what constitues "messy" code, it is clear that the code to uniformly close any open dialogs on a route change would a) be an extremely common/desirable use case, and b) could be considered boilerplate code.

The value of a framework/library is to eliminate this sort of boilerplate code.  So, while it could certainly be helpful to have a utility function to close all open dialogs - it would be more useful to have an option to set a default behavior, to call such a function on a route change.

Don

EisenbergEffect:
[...]

PR #218, opened 2016-10-08
feat(dialog-service): add .closeAll()

StrahilKazlachev:
Add a convenience method to close all open dialogs, see #214.

### DIFF
--- a/src/dialog-controller.js
+++ b/src/dialog-controller.js
@@ -76,11 +76,15 @@ export class DialogController {
             } else {
               this._reject(new DialogCancelError(output));
             }
-            return result;
+            return { wasCancelled: false };
           });
       }
 
       this._closePromise = undefined;
+      if (!this.settings.rejectOnCancel) {
+        return { wasCancelled: true };
+      }
+      return Promise.reject(new DialogCancelError());
     }, e => {
       this._closePromise = undefined;
       return Promise.reject(e);
--- a/test/unit/dialog-controller.spec.js
+++ b/test/unit/dialog-controller.spec.js
@@ -2,12 +2,10 @@ import {DialogController} from '../../src/dialog-controller';
 import {Container} from 'aurelia-dependency-injection';
 
 describe('the Dialog Controller', function () {
-  let dialogController;
   let renderer = {
     defaultSettings: {},
     hideDialog: Function.prototype
   };
-  let settings;
 
   function failOnRejection(promise, done) {
     promise.catch((reason) => {
@@ -16,18 +14,22 @@ describe('the Dialog Controller', function () {
     });
   }
 
-  beforeEach(function () {
-    new Container().makeGlobal();
-    settings = { name: 'Test' };
-    dialogController = new DialogController(renderer, settings, Function.prototype, Function.prototype);
+  function createDialogController(settings = {}) {
+    const dialogController = new DialogController(renderer, settings, Function.prototype, Function.prototype);
     dialogController.viewModel = {};
     dialogController.controller = {
       unbind() { }
     };
     dialogController.renderer.hideDialog = Function.prototype;
+    return dialogController;
+  }
+
+  beforeEach(function () {
+    this.dialogController = createDialogController();
   });
 
   it('should call close with success when ok method called', function () {
+    const dialogController = this.dialogController;
     let calledValue = 'Worked';
     spyOn(dialogController, 'close');
     dialogController.ok(calledValue);
@@ -35,6 +37,7 @@ describe('the Dialog Controller', function () {
   });
 
   it('should call close without success when cancel method called', function () {
+    const dialogController = this.dialogController;
     let calledValue = 'Didnt work';
     spyOn(dialogController, 'close');
     dialogController.cancel(calledValue);
@@ -42,6 +45,7 @@ describe('the Dialog Controller', function () {
   });
 
   it('should not be blocked from being closed if ".canDeactivate()" returns "false" once', function (done) {
+    const dialogController = this.dialogController;
     dialogController.viewModel = {
       canDeactivateCalls: 0,
       canDeactivate() {
@@ -60,40 +64,48 @@ describe('the Dialog Controller', function () {
     });
   });
 
-  describe('".close" method should', function () {
+  describe('".close" method should be', function () {
     beforeEach(function () {
+      this.dialogController = createDialogController();
       this.expectedOutput = {};
     });
 
-    it('resolve on success', function (done) {
-      spyOn(dialogController, '_resolve');
+    it('resolved on success', function (done) {
+      const dialogController = this.dialogController;
+      spyOn(dialogController, '_resolve').and.callFake((result) => {
+        expect(result.wasCancelled).toBe(false);
+        expect(result.output).toBe(this.expectedOutput);
+      });
       spyOn(dialogController, '_reject');
-      const result = dialogController.close(true, this.expectedOutput)
-        .then((result) => {
-          expect(result.wasCancelled).toBe(false);
-          expect(result.output).toBe(this.expectedOutput);
-          expect(dialogController._resolve).toHaveBeenCalledWith(result);
+      const pendingClose = dialogController.close(true, this.expectedOutput)
+        .then((closeOperationResult) => {
+          expect(closeOperationResult.wasCancelled).toBe(false);
+          expect(dialogController._resolve).toHaveBeenCalled();
           expect(dialogController._reject).not.toHaveBeenCalled();
           done();
         });
-      failOnRejection(result, done);
+      failOnRejection(pendingClose, done);
     });
 
-    it('resolve on cancellation if "rejectOnCancel" is false', function (done) {
-      spyOn(dialogController, '_resolve');
+    it('resolved on cancellation if "rejectOnCancel" is false', function (done) {
+      const dialogController = this.dialogController;
+      spyOn(dialogController, '_resolve').and.callFake((result) => {
+        expect(result.wasCancelled).toBe(true);
+        expect(result.output).toBe(this.expectedOutput);
+      });
       spyOn(dialogController, '_reject');
-      const result = dialogController.close(false, this.expectedOutput)
-        .then((result) => {
-          expect(result.wasCancelled).toBe(true);
-          expect(result.output).toBe(this.expectedOutput);
-          expect(dialogController._resolve).toHaveBeenCalledWith(result);
+      const pendingClose = dialogController.close(false, this.expectedOutput)
+        .then((closeOperationResult) => {
+          expect(closeOperationResult.wasCancelled).toBe(false);
+          expect(dialogController._resolve).toHaveBeenCalled();
           expect(dialogController._reject).not.toHaveBeenCalled();
           done();
         });
-      failOnRejection(result, done);
+      failOnRejection(pendingClose, done);
     });
 
-    it('resolve on cancellation if "rejectOnCancel" is true', function (done) {
+    it('resolved on cancellation if "rejectOnCancel" is true', function (done) {
+      const dialogController = this.dialogController;
       dialogController.settings.rejectOnCancel = true;
       spyOn(dialogController, '_resolve');
       spyOn(dialogController, '_reject').and.callFake((reason) => {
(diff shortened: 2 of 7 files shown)


StrahilKazlachev:
For me a `.closeAll()` convenience method should cover the following points:
- just close all current dialogs - the caller is not interested in awaiting completion
- close all and await them to close

@PWKad @EisenbergEffect @don-bluelinegrid I'd like input on the following: what should be the behavior when awaiting all to close?
1. `Promise.all` and if there is any error the caller should handle it - but this will settle on first error
2. `.catch()` on all `.cancel()` calls and just log the errors
3. `.catch()` on all `.cancel()` calls and return the errors as results - the caller can inspect the results array for the errors if interested - **currently submitted impl**
4. add an optional `ignoreErrors` parameter - `.closeAll(ignoreErrors = false)`. When `false` it will behave like **1.**, if `true` **2.**
5. ...

test/unit/dialog-service.spec.js
```diff
@@ -227,4 +227,49 @@ describe('the Dialog Service', () => {
...
+
+  it('reports no active dialog after ".closeAll" has been invoked', (done) => {
+    const settings = { viewModel: TestElement };
+    const dialogsToOpen = 6;
+    let i = dialogsToOpen;
+    let controllersPromises = [];
+    let controllers;
+    let catchWasCalled = false;
+
+    expect(dialogService.hasActiveDialog).toBe(false);
```
PWKad: Semantics here but we can move this to a separate describe or if block to state that this property resolves to true regardless.  This tests the unit more than the integration as well.  Not a stickler here though so perhaps we can merge this first and I can help to refactor the test in a new PR to demonstrate.

test/unit/dialog-service.spec.js
```diff
@@ -227,4 +227,49 @@ describe('the Dialog Service', () => {
...
+      return dialogService.closeAll();
+    }).catch((reason => {
+      catchWasCalled = true;
+    })).then(() => {
+      expect(catchWasCalled).toBe(false);
+      if (catchWasCalled) {
+        return done();
+      }
+      controllers.forEach((controller) => {
+        expect(controller.cancel).toHaveBeenCalled();
```
PWKad: There is also a `toHaveBeenCalledTimes()` assertion which takes an integer which may be helpful here (ex to have been called 6 times)

PWKad: Hmm actually disregard this should only call once per controller if I am reading this correctly.

test/unit/dialog-service.spec.js
StrahilKazlachev: @PWKad  I'd like to clear the behavior of `.closeAll()` first, then I'll add more tests in own `describe`. Can you check my first comment?

test/unit/dialog-service.spec.js
StrahilKazlachev: Yes, this are different instances.

PWKad:
So the developer has set up each view-model currently with a `.then` and `.catch` to watch for the result and when you call `closeAll` each of those modal's are `cancel`led.  I think in that situation it would have been up to the original caller to watch for those changes, right?

StrahilKazlachev:
@PWKad The promise returned from `DialogController#cancel()` is not the same as the one returned from `DialogService#open()`.
I thought some more on this and `.cancel()` can be rejected:
- bug in `Renderer#hideDialog()` - I don't see what can be done in that case except logging.
- bug in the `DialogController#close()` - ☝️ 
- the VMs `.canDeactivate()` and `.deactivate()` are rejected - user code
- ...

So now I don't think `DialogService#closeAll()` should do anything special in case of rejection. But what to do when `.canDeactivate()` returns false? The dialog will not close. Do we want to force it to close - if yes, how? Or we just return the controllers that didn't manage to close and leave it to the caller of `.closeAll()` to decide.

PWKad:
@StrahilKazlachev I'd say force close it and if the developer wants to do it without regard to `canDeactivate` they can do so.  Keep it simple and if anyone wants to open an issue on making it behave differently we can evaluate then.  Thoughts?

StrahilKazlachev:
@PWKad Currently the simplest way is to just call `.cancel()`, but it will take into consideration the result of `.canDeactivate()` and there is no force close API. I'll ponder on this some more.

PWKad:
@StrahilKazlachev anything you want to review before merging this?

StrahilKazlachev:
@PWKad Actually I'm currently thinking of closing this PR or just leave it hang till some other issues/questions get cleared. The initial idea from #214 was to add a convenience method that closes all open dialogs. `dialogService.controllers.forEach(ctrl => ctrl.cancel());` is it that much code? Also you can decide whether to await all promises to settle or not(ignoring rejections or not). Also what if `.canDeactivate()` returns `false`? The dialog will not close and I can't think of a way to force close such dialog. Is an API addition needed?

don-bluelinegrid:
@StrahilKazlachev 

I was the original reporter of https://github.com/aurelia/dialog/issues/214, the motivator for this work item issue.

As noted in the report, this is a core piece of functionality that is absent from the aurelia-dialog implementation.  I currently have an application using a number of dialogs based on aurelia-dialog, and if a dialog opens and then the view changes (session expires so redirects to login; user clicks back button; users clicks other navigation element; etc.) we see the dialog persisting over the view.  This is clearly not correct or what's desired, is just not good.  This is a common case/use case.  Putting the iteration code referenced above into an application defeats the purpose of the module/library, since it puts the burden of boilerplate code to solve a common case, onto the application developer; not to mention the burden of dealing with the other issues for edge cases that have been commented about in this thread.  Rather than every developer-user of the module having to solve these, shouldn't the module solve these for us?  @EisenbergEffect seemed to be in favor of resolving this, for these reasons.  So, why backtrack now?  To bring the module to a level of completeness, there should be a straightforward mechanism for handling this case.  As @EisenbergEffect  mentioned, a closeAll() method would be a *first step* towards this.  A more elegant solution would be some sort of configuration that allowed for specifying an event (like RouteChange) to automatically close all dialog instances.

Don

StrahilKazlachev:
@don-bluelinegrid If I didn't want to do this I wouldn't have created the PR in the first place. The number one concern for me is the inability to force close a dialog(`.canDeactivate()` returning `false`). I don't want to extend the API with **somewhat** working method that covers a specific scenario. Also it will be very nice to give your input on:
- do you want to await all dialogs to close before continuing or you just want to call `.close` and that's it
- if you want to await all to close do you want to ignore rejections or not - something like Promise.all() vs Promise.all().finally()

don-bluelinegrid:
@StrahilKazlachev 

Thanks for the response.

Regarding the questions -

> do you want to await all dialogs to close before continuing or you just want to call .close and that's it

It would be most flexible to have both options.  Providing an await/Promise implementation would give the developer-user the choice of using the Promise to await, or just making the call without waiting.

> if you want to await all to close do you want to ignore rejections or not - something like Promise.all() vs Promise.all().finally()

From the consumer's perspective, it would be expected to make a single call like ```DialogService.closeAll()``` that returns a single Promise (without the need for a developer-user to be concerned with the inner workings and needing to use Promise.all).  And, for most flexibility, and to meet generally expectations of use of Promises, it would make sense to reject if something failed - if an open dialog did not actually close.  Again, the developer-user would have the option of if or how to handle such a rejection.  And, they should be able to use .finally() of the single Promise return by .closeAll().

Thanks,
Don

StrahilKazlachev:
@PWKad @EisenbergEffect I'm trying to work on this one, but there is something bothering me. The `DialogService` has an instance property `.controllers`(isn't very descriptive) and `.hasActiveDialog`(`hasActiveDialog = !!controllers.length`). What does `active` mean? Just `CompositionEngine.compose` has succeeded? Or `DialogRenderer.showDialog` has also succeeded? Currently a controller is added to `.controllers` after `.compose` succeeds and before `.showDialog()` is called. I'm asking this because currently you could take a controller from `.controllers` and try to close it before `.showDialog` has completed.

PWKad:
Some of these properties have hung on over time if you have any recommendations for cleaning up,  renaming,  or otherwise removing unused complexity I'm game provided it doesn't break the public api.   Even if it does if we target it for a v1 release I think that is acceptable.

StrahilKazlachev:
@PWKad I was thinking we are more interested if there are **open** dialogs. If a dialog succeeds `compositionEngine.compose` but fails `dialogRenderer.showDialog` it should not be added to `.controllers`(as currently implemented `.showDialog` should fail only if it throws - so it will be a bug) . I'll update the PR.
It won't be completed before #225 because of the new setting - `rejectOnCancel`, since we will have to filter out rejections from cancellation(when the setting is on).

EisenbergEffect:
@PWKad Where are we at with this?

PWKad:
@StrahilKazlachev Since this depends on #225 I'll wait until this can be rebased on top of that when it is merged in tomorrow (pending no issues) and we can review then.

PWKad:
How to manage the controllers that did not manage to close?

1. canDeactivate returns false?
  a. Current behavior - returns a promise that never resolves.
  b. Future behavior - return a `closeResult` with property to indicate the cancellation.

```
let promise = this.dialogController.close('true').then(result => {
  if (result.wasCancelled) {
    // was not closed
  }
});
```

2. Has error thrown on close (could be because deactivate threw `Error` or canDeactivate threw `Error`)?
  a. Return in the array of close dialog results that did not deactivate properly
  a. return a `closeResult` with property to indicate the cancellation.

StrahilKazlachev:
@PWKad ready for review.

1. `DialogController.prototype.close` refactored as discussed.
  - always returns `{wasCancelled: boolean}`
  - if `rejectOnCancel` is `true` the promise will be rejected with `CancelDialogError`
  - it is a breaking change since when the close was a success it was returning the dialog result - given that the result is constructed with the data passed by the caller I don't think this is a big problem.
2. `DialogService.prototype.closeAll`
  - returns an array with all controllers whose closing was cancelled
  - fails if any of the calls to "dialogController.cancel" fails.
    - open to discussion whether `.closeAll` should not fail at all and just return the controller, as not closed, when the call to `.cancel` fails and log and error
3. `DialogService` - changed adding to `controllers` to happen after `.showDialog` succeeds, if it fails just call `.deactivate` on the controller to do its clean-up. This way you get access to a controller only if it's open. And that's why I aliased `.hasActiveDialog` with `.hasOpenDialog`. If we keep this changes `.hasActiveDialog` should be removed for v1.
@don-bluelinegrid Any comments on 2?

PWKad:
@StrahilKazlachev if you don't mind can you re-commit without changes to `dist` folder?  We usually only add the changes to those files during release.

StrahilKazlachev:
@PWKad it's time to get the hang of it when to use `push -f` 😞

PWKad:
@StrahilKazlachev did you lose some work?

StrahilKazlachev:
@PWKad No, this should be all the intended changes. If you see something strange do tell.

PWKad:
No it's all good you just mentioned using force push

src/dialog-controller.js
```diff
@@ -76,11 +76,15 @@ export class DialogController {
             } else {
               this._reject(new DialogCancelError(output));
             }
-            return result;
+            return { wasCancelled: false };
```
PWKad: Should we be returning a result type here?  Probably doesn't matter much either way just curious (we may have talked about this before)

test/unit/dialog-controller.spec.js
```diff
@@ -16,32 +14,38 @@ describe('the Dialog Controller', function () {
...
     dialogController.viewModel = {};
     dialogController.controller = {
       unbind() { }
     };
     dialogController.renderer.hideDialog = Function.prototype;
+    return dialogController;
+  }
+
+  beforeEach(function () {
+    this.dialogController = createDialogController();
```
PWKad: What is `this.` in this context?  If the test, generally I think it's more common to use a variable in this context as opposed to tacking on to this, but I'll look at this further to make sure I'm not missing something.

test/unit/dialog-controller.spec.js
```diff
@@ -16,32 +14,38 @@ describe('the Dialog Controller', function () {
...
-    new Container().makeGlobal();
-    settings = { name: 'Test' };
-    dialogController = new DialogController(renderer, settings, Function.prototype, Function.prototype);
+  function createDialogController(settings = {}) {
+    const dialogController = new DialogController(renderer, settings, Function.prototype, Function.prototype);
     dialogController.viewModel = {};
     dialogController.controller = {
       unbind() { }
     };
     dialogController.renderer.hideDialog = Function.prototype;
```
PWKad: I just realized this was in the test (and I know it's not from your code or this PR), not sure when it was added but something feels wrong about this.  We can clean it up later but seems like a weird thing to do :)

test/unit/dialog-controller.spec.js
PWKad: Looks like this may be a way to get around defining the same variable twice, but naming should take care of this.

src/dialog-controller.js
StrahilKazlachev: Did it for consistency, return type like `{wasCancelled: boolean} | void` seemed strange. Also I thought it will not be an overhead since it should be done once by controller.

test/unit/dialog-controller.spec.js
StrahilKazlachev: I'm using it like [here](https://jasmine.github.io/2.5/introduction#section-The_%3Ccode%3Ethis%3C/code%3E_keyword).

>  , but naming should take care of this.

What should I change?

don-bluelinegrid:
@StrahilKazlachev 
RE: https://github.com/aurelia/dialog/pull/218#issuecomment-267788039
> @don-bluelinegrid Any comments on 2?

I've been following the entire thread of discussion on this, and there had been a large amount of discussion around failure/rejection, and how this relates to closing more than one dialog.  With that, I have a few high-level questions regarding the overall approach -

1. Under what circumstances would the be more than one dialog rendered in a given moment?  Is this allowed by the framework?

2. Much of the discussion was around handling failures/rejections from multiple controllers.  It seemed to me that this should be conceptually the same as handling a failure/rejection from a single controller.  When I look at the source code for DialogController.close(), I see that there is an error/rejection handler, however, it's not clear what would possibly cause such an error.  The actual close() method does not do very much, and the error handler is fairly vague in terms of what exception it is catching.  So, I'm wondering what case could cause such a rejection, and whether a complicated failure/rejection-handling strategy is necessary in the solution to the closeAll() case.

3. In terms of the semantics, I would expect that a caller of closeAll() basically expects any rendered dialogs to be closed, and if for some reason that doesn't happen, some informative error data structure should be returned.  However, I'm not clear on the value of returning the complex structure of an array of DialogController - the question is, if for some reason closeAll() is called, and not all dialogs do close, then what is the recovery strategy at that point?  Can the caller do something useful to recover from this?  If not, then I'm not clear what the usefulness would be of the returned data structure vs. some exception message or enum/key.

Thanks,
Don

test/unit/dialog-controller.spec.js
PWKad: Gotcha, typically we've been using it like this - 

```
import {DialogController} from 'src/dialog-controller';
import {Container} from 'aurelia-dependency-injection';
describe('ModuleName', () => {
  let dialogController;
  let container;
  beforeEach(() => {
    container = new Container();
    dialogController = container.get(DialogController);
  });
});
```

which is basically the same as what you have just a slight style difference.  It works either way so no biggie, just hadn't seen it that way before.

StrahilKazlachev:
@don-bluelinegrid 
1. There is no restriction how many dialogs are open at the same time. The last one opened is the visible one, when closed the previous one becomes visible - stack.pop.
2. The rejection handler in `DialogController.prototype.close` only does some clean-up, then _passes_ on the rejection. A rejection can come from the renderer `.hideDialog()` method, or in the view model `.canDeactivate()` or `.deactivate()`, in those cases the source of the rejection will most likely be a bug. Also if `.canDeactivate()` returns false **and** the new option `rejectOnCancel` is `true`, the promise will be rejected. I want to emphasize that the promises returned by `DialogService.prototype.open` and `DialogController.prototype.close` are _different_.
3. There is no recovery strategy if not all dialogs get closed or if any of the calls to `.close` is rejected. The caller of `.closeAll()` should handle it, that is why the controllers get returned. I'm against the idea to blindly _hide_ those whose call to `.close()` got rejected or cancelled. And if the renderer `.hideDialog()` got rejected there isn't a way to hide it.
There could be a `forceHide` that when set to `true` will call `controller.renderer.hideDialog()` for every that failed to close, skipping `.canDeactivate()` and `.deactivate()`. Now if `.hideDialog()` is rejected I don't see what to do. @PWKad what do you think about such option or any other comments as a whole.

PWKad:
Whoever opened the dialog would have the error returned to them still though, correct?  Or did I miss that point.

StrahilKazlachev:
@PWKad The who opened the dialog will get an error only in the following cases:
- a promise got rejected somewhere along the chain from calling `DialogService.prototype.open()` till `.showDialog()` - `.showDialog()` is the last in the chain that can cause rejection.
- explicitly by calling `.error()` on the `DialogController`
- when calling `.close(false)` on the `DialogController` and `rejectOnError` is `true` - this is the only new one

The `DialogController` resolves or rejects the promise returned from `.open()` only after the chain from `canDeactivate`, `deactivate` and `hideDialog` succeeds. `.error` skips calling `canDeactivate`.

PWKad:
@StrahilKazlachev Ready to merge?

StrahilKazlachev:
@PWKad I'm ready. We can always revisit any comments before we hit v1. Should I merge it?

PWKad:
Sure.  This is the process, typically - 

1. Pull down branch locally.
2. Run tests
3. Run linter
4. If everything looks good you can use the big green button on GitHub just use the rebase option instead of the merge if possible (just keeps history a bit cleaner)

Master has an issue right now with lint so I'm going to fix that up real fast and push up.

PWKad:
Woops realized you had fixed the lint issues so only removed the ones I created and pushed up, should be good to go now :)

don-bluelinegrid:
@EisenbergEffect @PWKad @StrahilKazlachev 

A couple of questions about this -

1. It looks like the addition of the closeAll() method was merged to master 18 days ago, in dialog-service.  However, there is no new tag/release that includes this, so it is effectively not available for use via npm.  What is the plan here to get this released to a tagged version?

2. When I look at dialog-service in master, I see that indeed there is a new closeAll() method.  However, I'm also noticing that the openAndYieldController() method is removed.  What's the background on this?  We are relying on this in our application.

Thanks,
Don

StrahilKazlachev:
About `.openAndYieldController()` see #224 and related issues/PR's - you can still get the controller.
We are working towards v1 and trying to make the API simpler and easier to use.
Also the dialog is being migrated to TS(more accurate typing), so it should be easier to migrate.
I'll see if we can put some kind of a roadmap.

PWKad:
@don-bluelinegrid The next release will include a series of breaking changes and will be the last beta prior to v1.  We will compile a list of those changes to include in the release and try to get the release out in the next week or so.

Once released this should stabilize all APIs and we'd love to have your feedback on the upgrade process before v1 to make sure nothing is missing and working as expected for all of your use cases.  Does that seem reasonable?

don-bluelinegrid:
@PWKad 

Patrick,

I'm checking again, and I see that the release tag is up to `1.0.0-beta-3.0.1`.  Yet, I don't see the inclusion of the `closeAll()` method in this tag or in master.

What's the status?  I would have expected a near-release beta to include all of the code planned for the release.  I'm eagerly waiting for this to be released in order to fix glaring UX problems when using dialogs.

Thanks,
Don

PWKad:
We have breaking changes queued up right now for the final beta release - as a note the beta 3.0.1 tag is from late October, I'll try to get a better exact date for final release and post it here.

don-bluelinegrid:
Status?

don-bluelinegrid:
@PWKad 

Patrick,

on Jan 29 you mentioned preparing for a final beta release.  Do you have status on this?  We have bugs in our product backlog related to the inability to close arbitrary dialogs on view change, which depends on this fix, and our next release is approaching - so we are trying to determine if these will be resolvable.

Thanks,
Don

PWKad:
@don-bluelinegrid sorry for the radio silence - we are wrapping up refactoring the final bits to TypeScript which is mostly complete save for a few tooling changes.  We are planning a meeting for Monday morning to finalize the tooling bits and then push up a release candidate.  I'll be sure to ping you as soon as it is ready.  Again, apologies for the delay.

PR #221, opened 2016-10-17
fix(lifecycle): fix lifecycle methods default value

StrahilKazlachev:
The default value was applied only for sync implementations. `Promise.resolve()` will also handle `PromiseLike` results.


PWKad:
LGTM is there any corresponding issue or is this just clean-up?  Merging either way ;)

StrahilKazlachev:
@PWKad No, there wasn't.

Issue #224, opened 2016-10-19
Consolidate .open() and .openAndYieldController()


StrahilKazlachev:
**I'm submitting a feature request**
- **Library Version:**
  1.0.0-beta.3.0.1

**Please tell us about your environment:**
- **Operating System:**
  Windows [7|10]
- **Node Version:**
  6.2.0
- **NPM Version:**
  3.8.9
- **JSPM OR Webpack AND Version**
  JSPM 0.16.32
- **Browser:**
  all
- **Language:**
  all

**Current behavior:**
Two methods to open a dialog:
- `.open(settings: DialogSettings): Promise<DialogResult>`
- `.openAndYieldController(setings: DialogSettings): Promise<DialogController & { result: Promise<DialogResult> }>`

**Expected/desired behavior:**
- **What is the expected behavior?**
  To consolidate the two methods as follows:
  - **rename** `DialogResult` => `CloseDialogResult`
  - add **new** dialog result interface

```
interface OpenDialogResult { 
  controller: DialogController | null;
  wasCancelled: boolean;
  result: Promise<CloseDialogResult>
}
```
- the merged `.open()` definition 

```
open(settings: DialogSettings): 
        { openResult: Promise<OpenDialogResult>, closeResult: Promise<CloseDialogResult> };
// OR
open(settings: DialogSettings & { yieldController: false | undefined }): Promise<CloseDialogResult>;
open(settings: DialogSettings & { yieldController: true }): Promise<OpenDialogResult>;
open(settings: DialogSettings & { yieldController: boolean | undefined }): Promise<OpenDialogResult | CloseDialogResult>;
```
- **What is the motivation / use case for changing the behavior?**
  First of all, I get that `openAndYieldController` was introduced so that new functionality could be added in non-breaking way, but
  - confusion when to use `.open()` and when `.openAndYieldController()` - I think there was at least 1 issue, also on the gitter channel, and at work(I shouldn't have said that).
  - adding the `result: Promise<DialogResult>` property on `DialogController` doesn't feel right, also it's somewhat hard to type currently #216
  - with the current impl `.open()` is a subset of `.openAndYieldController()`
  - in the context of #223 we can propagate cancellation in consistent way - having `.wasCancelled` in both `DialogOpenResult` and `DialogCloseResult`(cancelable promises do seem far 😞)

I'll make a PR to comment on actual code if these changes will be considered.

EisenbergEffect:
[...]

Issue #223, opened 2016-10-19
.open() never resolves if canActivate() returns false


StrahilKazlachev:
**I'm submitting a bug report**
- **Library Version:**
  1.0.0-beta.3.0.1

**Please tell us about your environment:**
- **Operating System:**
  Windows [7|10]
- **Node Version:**
  6.2.0
- **NPM Version:**
  3.8.9
- **JSPM OR Webpack AND Version**
  JSPM 0.16.32
- **Browser:**
  all
- **Language:**
  all

**Current behavior:**
If the VM `.canActivate()` method returns `false` the whole part where `compositionEngine.compose()` and `renderer.showDialog()` is skipped. If you have called:
- `.open()` - the returned promise will never resolve.
- `.openAndYieldController()` - the returned promise will resolve with the controller, but the `.result` property added to the controller will never resolve.

**Expected/desired behavior:**
- **What is the expected behavior?**
  Since there is no retry mechanism if `.canActivate()` returns false that should be considered a cancellation. The caller of `.open()`/`openAndYieldController()` should be notified for the cancellation or inability to proceed, if returning `false` is not considered a (early) cancellation.

mattbeckman:
I think I was bit by this bug today. I couldn't get the dialog promise to return after using controller.ok().

I bypassed by using a canDeactivate() method to do a delayed promise resolution:

`    canDeactivate() {
        return new Promise(function(resolve, reject) { 
            console.log("Called canDeactivate");
            window.setTimeout(
                function() {
                    resolve(true);
                }, 100);
        });
    }`

davismj:
@pwkad Do we think the promise returned from `.open` should `reject` or `resolve(false)`?

PWKad:
@davismj `reject` only if the developer asks for it specifically to keep a good dev experience

StrahilKazlachev:
Closed #225.

Issue #217, opened 2016-10-08
Convert to src to TypeScript


PWKad:
Right now we have recognized a few issues in the build process such as missing the ability to support union types and some of the other ts2.0 new type features.  The build process also can be a bit tricky since there are so many moving parts.  With Aurelia-Validation I believe @jdanyow and @niieani have done a good job of showing that the plugin repo's can use TS in a nice, clean, and elegant fashion.

I'm proposing we convert over the existing codebase to use the TypeScript build and enforce typings a bit further.

Pros:

Easier build process
Easier to enforce typings are correct

Cons: 

Could be intimidating to ask for PRs if the developer isn't comfortable with TypeScript
Adds no real "features" to the repo.

@EisenbergEffect @StrahilKazlachev @cmichaelgraham and anyone else mentioned or who is interested feedback welcome.

EisenbergEffect:
I'm ok with this but we need a full working build process. The Validation library is actually in TypeScript but it's kind of a pain because it's missing a lot of features that the build from our other libraries have. We need to get that library at 100% fidelity first and then we can use it as a template to move other libraries over.

jdanyow:
https://github.com/aurelia/validation/issues/357

EisenbergEffect:
Thanks!

PWKad:
Agreed then that those issues need to be resolved first.  I'll comment there.

PWKad:
Still waiting on a few last tasks before converting just giving an update

PWKad:
@StrahilKazlachev I think we can safely start on this after the current PR's are accepted as all of the outstanding issues from validation are handled (except for one last task which has a PR that will be merged this week)

If you want to schedule a quick call some time over the next week or two to discuss current state and how we can do this whilst achieving the coveted v1 tag let me know!  Thanks again for all your contributions, they do not go unnoticed.

PWKad:
For the initial implement -

1. Convert build.js task to use `gulp-typescript` instead of babel
2. Remove dts-generator
3. Make sure .d.ts is being generated / output
4. Make sure output matches current output structure

After that we can review tooling.

niieani:
@PWKad I think it might be easier to just switch to the stack used by any of the Aurelia repos which already use TypeScript, and get rid of gulp dependencies.

PWKad:
After converting over we will probably do that, trying to isolate to one major change at a time.

StrahilKazlachev:
Done and released. Common build infrastructure is being discussed separately.

PR #225, opened 2016-10-20
feat(dialog-service): refactor consolidate open methods in one

StrahilKazlachev:
Starting point for #224, #223 
Will work on the tests after we clear the new open API.

### DIFF
--- a/src/dialog-service.js
+++ b/src/dialog-service.js
@@ -4,8 +4,8 @@ import {CompositionEngine, ViewSlot} from 'aurelia-templating';
 import {DialogController} from './dialog-controller';
 import {Renderer} from './renderer';
 import {invokeLifecycle} from './lifecycle';
-import {DialogResult} from './dialog-result';
 import {dialogOptions} from './dialog-options';
+import {DialogSettings, OpenDialogResult, CloseDialogResult} from './interfaces';
 
 /**
  * A service allowing for the creation of dialogs.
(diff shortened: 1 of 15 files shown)


StrahilKazlachev:
@EisenbergEffect @PWKad What do you think about the `.open` method API? Is returning a structure like `{ openResult: Promise<OpenDialogResult>, closeResult: Promise<CloseDialogResult> }` too unfriendly? Do you prefer an API where you pass a setting to control the output:
```
open(settings: DialogSettings & { yieldController: false | undefined }): Promise<CloseDialogResult>;
open(settings: DialogSettings & { yieldController: true }): Promise<OpenDialogResult>;
```
Or you have something else?

PWKad:
I was trying to find where I had mentioned this before but combing back through issues and gitter I couldn't.  I prefer the second that you had brought up before.

So the usage would be something like - 

```
export class MyViewModel {
  submit(){
    this.dialogService.open({ viewModel: Prompt, model: 'Good or Bad?'}).then(response => {
      if (!response.wasCancelled) { console.log('good'); }
    });
  }
}
```
And with the second flag
```
export class MyViewModel {
  submit(){
    this.dialogService.open({ viewModel: Prompt, model: 'Good or Bad?', yieldController: true }).then(controller => {
      controller.result.then(response => {
        if (!response.wasCancelled) { console.log('good'); }
      }
    });
  }
}
```
Right?

StrahilKazlachev:
@PWKad On gitter, but I was left with the impression that you preferred the other solution, good thing I asked again 😅 For this signature though we will need #217 to properly type it, at least I don't see how to do the overloads with the current generator.
I'll write back more questions after pushing the changes.

StrahilKazlachev:
@PWKad Updated the `.open()` API as discussed + tests. A couple of questions:
* naming
 * new interfaces: `OpenDialogResult`, `CloseDialogResult` - comments?
 * `OpenDialogResult#closeResult` or just `result`(too much **result** 😕 ). Will it be confusing/misleading if we name it `output` also(`DialogResult`/`CloseDialogResult`)?
* can you comment on #223 because I've designed `OpenDialogResult`(`#wasCancelled`) so that `.canActivate() => false` is considered a cancellation.

src/dialog-service.js
```diff
@@ -34,32 +34,54 @@ export class DialogService {
...
+
+    closeResult.then(() => {
       _removeController(this, dialogController);
     }, () => {
       _removeController(this, dialogController);
     });
-    return _openDialog(this, childContainer, dialogController)
-      .then(() => dialogController);
+
+    let openResult = _getViewModel(this.container, this.compositionEngine, childContainer, dialogController).then((instruction) => {
```
PWKad: I think there should be a task to somehow reduce the complexity in here a bit later down the road but looks good for now!

PWKad:
@StrahilKazlachev anything we need to answer before merging this?

StrahilKazlachev:
@PWKad Will add some more tests, update readme and sample/example.

test/unit/dialog-renderer.spec.js
```diff
@@ -5,7 +5,7 @@ import {dialogOptions} from '../../src/dialog-options';
 import {configure} from '../../src/aurelia-dialog';
 import {TestElement} from '../fixtures/test-element';
 
-describe('the Dialog Renderer', () => {
+describe('the Dialog Renderer', function () {
```
PWKad: Just curious if there was any reason to change these over?

StrahilKazlachev: `describe` and `it` set the context - `this`, using a lambda prevents it. If someone tries to use `this` without noticing a lambda has been used, he(me :sweat_smile:) can loose quite some time.

PWKad:
Going to merge this first thing in the morning, please let me know if there is an order to merge in.

StrahilKazlachev:
@PWKad no order for this one.

PR #230, opened 2016-11-06
fix(lefecycle): handle thrown errors

StrahilKazlachev:
reject the promise chain with the error thrown by the invoked lifecycle method

### DIFF
--- a/src/lifecycle.js
+++ b/src/lifecycle.js
@@ -8,7 +8,9 @@
  */
 export function invokeLifecycle(instance: any, name: string, model: any) {
   if (typeof instance[name] === 'function') {
-    return Promise.resolve(instance[name](model)).then(function(result) {
+    return Promise.resolve().then(() => {
+      return instance[name](model);
+    }).then(function (result) {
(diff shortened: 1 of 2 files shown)


davismj:
I'm not sure what to say about this. Was this not working as intended? At first glance, swallowing errors seems like a bad idea.

PWKad:
This doesn't swallow errors, it propagates them back out to the original caller

StrahilKazlachev:
Yes, it propagates them since calling the lefecycle method is done in `.then(() => {...})`. I prefer this to try/catch. I missed this case in my previous PR.

CasiOo:
Will this resolve the issue where you get exception messages such as `h.load() is not a function`?
We experienced the above exception in a CLI project when the view-model existed but the view file had a spelling mistake in the name (it couldn't find the view).

StrahilKazlachev:
@CasiOo I would say no. This should resolve any problems related to `canActivate/canDeactivate/deactivate` throwing an error.

PR #236, opened 2016-11-19
fix(dialog-renderer): check for transitions before awaiting transitonend

### LINKED ISSUE #190: "position" handler and no transitions issue

StrahilKazlachev:
fix #190

### DIFF
--- a/src/dialog-renderer.js
+++ b/src/dialog-renderer.js
@@ -25,26 +25,20 @@ let transitionEvent = (function() {
     }
   };
 }());
-let hasTransition = (function () {
-  const defaultDuration = '0s';
-  let transitionDuration;
-  let t;
-  let el = DOM.createElement('fakeelement');
-  let transitions = {
-    'transition': 'transitionDuration',
-    'OTransition': 'oTransitionDuration',
-    'MozTransition': 'transitionDuration',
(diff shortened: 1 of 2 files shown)


src/dialog-renderer.js
```diff
@@ -25,6 +25,28 @@ let transitionEvent = (function() {
     }
   };
 }());
+let hasTransition = (function () {
+  const defaultDuration = '0s';
+  let transitionDuration;
+  let t;
+  let el = DOM.createElement('fakeelement');
+  let transitions = {
+    'transition': 'transitionDuration',
```
StrahilKazlachev: @PWKad I'm not entirely sure about prefixed names.

PWKad:
@StrahilKazlachev this looks good to me, can you let me know if there is an order to merge in at?  As far as your question on `is this the best way to check for transitions in flight?` it seems so to me but perhaps @zewa666 could take a quick peek and see based on his previous work with multiple animation libraries if we need to account for any edge cases?

StrahilKazlachev:
@PWKad no order for this one.

Issue #238, opened 2016-11-26
FR: Add ability to enable esc without `lock: false`


PWKad:
See #234 -

Add an option on creating of dialog to allow esc key without adding the close button and such.

PR #240, opened 2016-12-13
feat(dialog options): allow esc to close dialog without lock: false

devanp92:
Adding ability to enable esc without `lock:false`.

Referenced: https://github.com/aurelia/dialog/issues/238

@PWKad

### DIFF
--- a/src/dialog-options.js
+++ b/src/dialog-options.js
@@ -4,5 +4,6 @@ export let dialogOptions = {
   startingZIndex: 1000,
   ignoreTransitions: false,
   rejectOnCancel: false,
-  yieldController: false
+  yieldController: false,
+  enableEscClose: true
 };
--- a/src/dialog-renderer.js
+++ b/src/dialog-renderer.js
@@ -47,7 +47,7 @@ export class DialogRenderer {
   _escapeKeyEventHandler = (e) => {
     if (e.keyCode === 27) {
       let top = this._dialogControllers[this._dialogControllers.length - 1];
-      if (top && top.settings.lock !== true) {
+      if (top && (top.settings.lock !== true && !top.settings.enableEscClose)) {
         top.cancel();
       }
     }
--- a/src/dialog-renderer.js
+++ b/src/dialog-renderer.js
@@ -47,7 +47,7 @@ export class DialogRenderer {
   _escapeKeyEventHandler = (e) => {
     if (e.keyCode === 27) {
       let top = this._dialogControllers[this._dialogControllers.length - 1];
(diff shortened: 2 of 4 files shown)


PWKad:
Nice, thanks for the contribution!  Going to pull down in a second and test it out.

src/dialog-renderer.js
```diff
@@ -47,7 +47,7 @@ export class DialogRenderer {
   _escapeKeyEventHandler = (e) => {
     if (e.keyCode === 27) {
       let top = this._dialogControllers[this._dialogControllers.length - 1];
-      if (top && top.settings.lock !== true) {
+      if (top && (top.settings.lock !== true && !top.settings.enableEscClose)) {
```
PWKad: I think this should be `!top.settings.enableEscClose` because when I pull this down and test it in the sample app it does not close, when I set `enableEscClose: false` it does, which seems backwards.

StrahilKazlachev:
I was thinking that we need something more flexible. Like for starters adding a new `keyboard` setting - `keyboard: boolean | ("ESC | "ENTER"")[]`, and later extend it to support custom keycode-handler pairs.
- `keyboard: true` - like now - `dialogController.close()`
- `keyboard: ("ESC" | "ENTER")[]`
  - on ESC - `dialogController.close()` and if `keyboard` contains `"ESC"`
  - on ENTER - `dialogController.ok()` and if `keyboard` contains `"ENTER"`
I think that for simple confirmation dialogs having the option to also wire-up the `ENTER` will be great. Also I've had a requirement(in Durandal app) to be able to dismiss simple notification-like dialog(only with "OK"/"CLOSE" button) both with ESC and ENTER.

devanp92:
@PWKad Updated. Also updated README.md.

PWKad:
@StrahilKazlachev What do you think about merging this in silently (not showing in docs) and adding a todo to move to keyboard mappings?

StrahilKazlachev:
@PWKad I don't mind. Just mentioned it since the `lock` setting has been bugging me for a while - controlling 3 different things with just 2 states.

src/dialog-options.js
```diff
@@ -4,5 +4,6 @@ export let dialogOptions = {
   startingZIndex: 1000,
   ignoreTransitions: false,
   rejectOnCancel: false,
-  yieldController: false
+  yieldController: false,
+  enableEscClose: true
```
PWKad: I think we should default to `false` initially, thoughts?  Since lock is default true this overrides that behavior by default.  I think we should make this conventionally false and configurably true.

PR #255, opened 2017-03-03
feat(dialog-controller): add ok flag as parameter to canDeactivate

ydbondt:
When implementing canDeactivate there is no way to see if the user signified success.
Added ok flag as a param to canDeactivate.

### DIFF
--- a/src/dialog-controller.js
+++ b/src/dialog-controller.js
@@ -63,7 +63,7 @@ export class DialogController {
       return this._closePromise;
     }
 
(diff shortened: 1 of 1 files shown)


PWKad:
Hi thanks for this PR - can you describe what isn't working currently in a bit more detail?  We are currently refactoring this code so there is going to be a possibility to get this in just need to know a bit more details, perhaps opening an issue and then connecting it would help to understand a bit better what is missing.

PR #259, opened 2017-03-08
feat(all): decompose "lock" setting

StrahilKazlachev:
~~remove "lock" setting~~
add "keyboard" setting
add "backdropDismiss" setting
add "showCloseButton" bindable property to ai-dialog-header

### DIFF
--- a/src/ai-dialog-body.ts
+++ b/src/ai-dialog-body.ts
@@ -1,4 +1,4 @@
-import {customElement, inlineView} from 'aurelia-templating';
+import { customElement, inlineView } from 'aurelia-templating';
 
 @customElement('ai-dialog-body')
 @inlineView(`
--- a/src/ai-dialog-footer.ts
+++ b/src/ai-dialog-footer.ts
@@ -1,5 +1,5 @@
-import {customElement, bindable, inlineView} from 'aurelia-templating';
-import {DialogController} from './dialog-controller';
+import { customElement, bindable, inlineView } from 'aurelia-templating';
+import { DialogController } from './dialog-controller';
 
 /**
  * View-model for footer of Dialog.
@@ -27,6 +27,9 @@ export class AiDialogFooter {
   @bindable public buttons: any[] = [];
   @bindable public useDefaultButtons: boolean = false;
 
+  /**
+   * @internal
+   */
   public static inject = [DialogController];
   constructor(public controller: DialogController) { }
 
--- a/src/ai-dialog-header.ts
+++ b/src/ai-dialog-header.ts
@@ -1,5 +1,5 @@
-import {customElement, inlineView} from 'aurelia-templating';
-import {DialogController} from './dialog-controller';
+import { customElement, bindable, inlineView } from 'aurelia-templating';
+import { DialogController } from './dialog-controller';
 
 @customElement('ai-dialog-header')
 @inlineView(`
@@ -8,7 +8,7 @@ import {DialogController} from './dialog-controller';
       type="button"
       class="dialog-close"
       aria-label="Close"
-      if.bind="!controller.settings.lock"
+      show.bind="showCloseButton"
       click.trigger="controller.cancel()">
       <span aria-hidden="true">&times;</span>
     </button>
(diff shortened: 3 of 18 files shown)


src/ai-dialog-body.ts
```diff
@@ -1,4 +1,4 @@
-import {customElement, inlineView} from 'aurelia-templating';
+import { customElement, inlineView } from 'aurelia-templating';
```
PWKad: Is the linter set up to enforce spacing here after open and before close brackets?

src/ai-dialog-footer.ts
```diff
@@ -27,6 +27,9 @@ export class AiDialogFooter {
   @bindable public buttons: any[] = [];
   @bindable public useDefaultButtons: boolean = false;
 
+  /**
+   * @internal
```
PWKad: Better to simply make `inject` private since it has no bearing on the usage, only on the output API?

src/ai-dialog-header.ts
```diff
@@ -8,7 +8,7 @@ import {DialogController} from './dialog-controller';
       type="button"
       class="dialog-close"
       aria-label="Close"
-      if.bind="!controller.settings.lock"
+      show.bind="showCloseButton"
```
PWKad: Generally I prefer using `if` over `show` unless there are descendants as it keep the DOM a bit cleaner but works the same either way just throwing it out there.

src/dialog-renderer.ts
```diff
@@ -128,7 +144,7 @@ export class DialogRenderer implements Renderer {
   private setupClickHandling(dialogController: DialogController): void {
     this.stopPropagation = e => { e._aureliaDialogHostClicked = true; };
     this.closeDialogClick = e => {
-      if (!dialogController.settings.lock && !e._aureliaDialogHostClicked) {
+      if (dialogController.settings.backdropDismiss && !e._aureliaDialogHostClicked) {
```
PWKad: For the setting name we discussed `dismissableOverlay`, I'm good with any name might be worth polling in the aurelia channel to the core team about the name since this should be RC-worthy

src/dialog-service.ts
```diff
@@ -185,7 +186,7 @@ export class DialogService {
         }
         return Promise.reject(reason);
       });
-    })).then((unclosedControllers) => unclosedControllers.filter(unclosed => !!unclosed));
+    })).then(unclosedControllers => unclosedControllers.filter(unclosed => !!unclosed));
```
PWKad: Sorry I missed this before but what is this actually filtering?  Seems to be checking that an item in an iterable array is not null, correct?

src/ai-dialog-body.ts
StrahilKazlachev: No, just followed `aurelia-validation`.

src/ai-dialog-footer.ts
StrahilKazlachev: I'm OK with that. My reasoning is that this is an API that will be consumed by something external - the DI `Container`. At the same time it is not intended for use by user code.

src/ai-dialog-header.ts
StrahilKazlachev: I don't mind also.

src/dialog-renderer.ts
StrahilKazlachev: `backdrop` - I went over some other dialogs/modals impls and this word was used
`dismissableOverlay` - for me this conveys that the overlay can be dismissed

src/dialog-service.ts
StrahilKazlachev: Yes. When a dialog successfully closes - `undefined` is returned, those that were cancelled return the controller itself. The end result should be and Promise<DialogController[]> - the controllers of the dialogs that were cancelled.

StrahilKazlachev:
@PWKad 
* `backdropDismiss` => `overlayDismiss` -in sync with out terminology
* `public static inject` - we have `"noUnusedLocals": true` and making it `private` gives compiler error
* added back `lock` - the individual settings - `keyboard`, `overlayDismiss`, override it

PR #265, opened 2017-03-22
feat: host and childContainer settings

### LINKED ISSUE #179: dialogs do not show up if browser is in  element fullscreen mode.
Need to add configurability, where should dialog attach with a default of 'body'.
So one could configure dialog (on per launch base) where to attach to.

### LINKED ISSUE #220: [feature] Ability to define custom renderer per-modal
- **Library Version:**
  1.0.0-beta.3.0.0

- **Operating System:**
  Windows 10
- **Node Version:**
  5.10.0
  
  
- **NPM Version:**
  3.10.7
  
  
- **JSPM OR Webpack AND Version**
  requirejs
  
  
- **Browser:**
  all | Chrome XX | Firefox XX | Edge XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView
- **Language:**
  TypeScript 2.0.2
- **What is the expected behavior?**
  Currently i'm faced with the problem of defining:
  1 default renderer
  2 renderer for notifications (non-blocking and with different behavior)

Thought of having smth like this:
dialogService.useRenderer(NotifyCustomRenderer);
- **What is the motivation / use case for changing the behavior?**
  Ability to use differently styled and behaved modals - same as aurelia-validation

### LINKED ISSUE #260: Support dialog integration in custom elements

* **Library Version:**
1.0.0-beta.3.0.1

* **Operating System:**
all

* **Node Version:**
6.2.0

* **NPM Version:**
3.8.9

* **JSPM OR Webpack AND Version**
all

* **Browser:**
all

* **Language:**
all

**Current behavior:**
Can not make a dialog be part of a custom element - both DOM and DI hierarchy.
  1. the dialog has its Container created from the root Container
  2. can not specify where to append it in the DOM - always in the body #179

* **What is the expected behavior?**
To be able to specify the dialog host element and provide a child container to use instead of creating new one from the root container. Allowing for a Container provision could also solve #220

* **What is the motivation / use case for changing the behavior?**
Lets say we have a complex value we want to provide an editor for(custom element). That editor has a simplified and advanced mode(requires a lot of screen estate). The advanced mode is shown in a dialog. I want to seamlessly integrate that dialog as part of the custom element both DOM and DI hierarchy.
The complex value may be recursive in terms of type - it may contain a value, as a property - direct or not, that is of the same complex type(linked list like).

StrahilKazlachev:
closes #260, closes #179, closes #220
* `host: Element` - if provided the `Renderer` will attach the dialog to this element, defaults to the `body`
* `childContainer: Container`
  * the container in which the `DialogController` is registered and from which the `Renderer` is acquired
  * if not provided a child container is created from the root `Container`

### DIFF
--- a/src/dialog-renderer.ts
+++ b/src/dialog-renderer.ts
@@ -97,7 +97,19 @@ export class DialogRenderer implements Renderer {
 
   public dialogContainer: HTMLElement;
   public dialogOverlay: HTMLElement;
+  public host: Element;
   public anchor: Element;
+
+  private getOwnElements(parent: Element, selector: string): Element[] {
+    const elements = parent.querySelectorAll(selector);
+    const own: Element[] = [];
+    for (let i = 0; i < elements.length; i++) {
+      if (elements[i].parentElement === parent) {
+        own.push(elements[i]);
+      }
+    }
+    return own;
+  }
 
   private attach(dialogController: DialogController): void {
     const spacingWrapper = DOM.createElement('div'); // TODO: check if redundant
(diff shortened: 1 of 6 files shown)


StrahilKazlachev:
Not sure whether `childContainer` should be just `container`, thoughts?

EisenbergEffect:
Naming it childContainer is good and follows the same pattern as the composition engine. Feel free to merge this at will, based on your knowledge of the other PRs and TS work you are doing.

Issue #264, opened 2017-03-22
Invalid value for padding


glen-84:
`aurelia-dialog@1.0.0-beta.3.0.1`

The padding is set to a value of `3` (it is missing the unit of measurement).

PR #269, opened 2017-03-25
fix(inputs.less): fix invalid padding value

devanp92:
Resolves https://github.com/aurelia/dialog/issues/264


StrahilKazlachev:
* currently the `.less` file is not used - the css is inlined in `src/dialog-configuration.ts`
* all PRs currently must be made against the `rc-1.0.0` branch

PR #271, opened 2017-03-28
feat(dialog-controller): add object to canDeactivate / deactivate functions.

### LINKED ISSUE #261: feat(dialog-controller): Add indication of user action in canDeactivate.

* **Library Version:**
1.0.0-rc-1

* **Operating System:**
all

* **Node Version:**
6.2.0

* **NPM Version:**
3.8.9

* **JSPM OR Webpack AND Version**
JSPM 0.16.32

* **Browser:**
all

* **Language:**
all

**Current behavior:**
`canDeactivate` and `deactivate` function is called just before closing a dialog. However, it's not possible to see if the user has pressed OK, or just closed (or cancelled) the dialog. 

* **What is the expected behavior?**
Both functions should have parameter of type `DialogCloseResult`.

* **What is the motivation / use case for changing the behavior?**
1. You need to know whether the close was cancel or not, and/or the output, in `.canDeactivate()`/`.deactivate()` of the dialog VM.
2. You have passed the `DialogController` to some other API and the close was triggered programmatically through the `DialogController` API.
If we go with this API extension it should:

* cover both `.canDeactivate()` and `.deactivate()`
* pass both the `wasCancelled` flag and the output - could reuse the `DialogCloseResult` interface
* the changes must be done against the `rc-1.0.0` branch

ydbondt:
Resolves #261

### DIFF
--- a/src/dialog-controller.ts
+++ b/src/dialog-controller.ts
@@ -45,8 +45,8 @@ export class DialogController {
   /**
    * @internal
    */
-  public releaseResources(): Promise<void> {
-    return invokeLifecycle(this.controller.viewModel || {}, 'deactivate')
+  public releaseResources(dialogResult?: any): Promise<void> {
+    return invokeLifecycle(this.controller.viewModel || {}, 'deactivate', dialogResult)
       .then(() => this.renderer.hideDialog(this))
       .then(() => { this.controller.unbind(); });
   }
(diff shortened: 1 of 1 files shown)


src/dialog-controller.ts
```diff
@@ -45,8 +45,8 @@ export class DialogController {
   /**
    * @internal
    */
-  public releaseResources(): Promise<void> {
-    return invokeLifecycle(this.controller.viewModel || {}, 'deactivate')
+  public releaseResources(dialogResult? :any): Promise<void> {
```
StrahilKazlachev: fix lint error

src/dialog-controller.ts
```diff
@@ -97,17 +97,20 @@ export class DialogController {
...
 
-    return this.closePromise = invokeLifecycle(this.controller.viewModel || {}, 'canDeactivate').catch(reason => {
-      this.closePromise = undefined;
-      return Promise.reject(reason);
+    const dialogResult = { wasCancelled: !ok, output } as DialogCloseResult;
+
+    return this.closePromise = invokeLifecycle(this.controller.viewModel || {}, 'canDeactivate', dialogResult)
+      .catch(reason => {
+        this.closePromise = undefined;
+        return Promise.reject(reason);
```
StrahilKazlachev: fix indentation

StrahilKazlachev:
@PWKad One more opinion will be appreciated.
One other thing that should be handled is the `.error` method.
* it does call only the  `.deactivate` method
* currenlty it is invoked with a `message: any`
* this is not currently provided to the `.deactivate` method

We should make it possible to determine in the `.deactivate` method whether `.error` was called.
I think that we should either type the `message` param as `message: Error` or leave it `message: any` and create new dialog error with a `reason` property that will hold the message. Thoughts and other suggestions?

src/dialog-controller.ts
ydbondt: Stupid question, but how do you expect the indentation to be ? :blush: 
Reason for splitup is the 200 chars limitation.

ydbondt: How do you expect the indentation to be ? :blush:

ydbondt:
@StrahilKazlachev Done

StrahilKazlachev:
@ydbondt  Can you give your input on https://github.com/aurelia/dialog/pull/271#issuecomment-290433121?

PR #272, opened 2017-03-31
fix(dialog-configuration): use PLATFORM.moduleName

### LINKED ISSUE #254: Compatibility with new webpack build
One goal of the new webpack build is that libs work out-of-the-box with zero config.
To achieve this goal a few small changes are required.
- [x] Be sure to wrap dynamically loaded modules into `PLATFORM.moduleName`, for example here: https://github.com/aurelia/dialog/blob/master/src/dialog-configuration.js#L9-L13
- [ ] @niieani mentionned that this lib encourages passing a VM loaded with a static `import` and then asking Aurelia to find a matching view, which requires hacky origin stuff. I have added support for that scenario in the webpack plugin, but we need to verify that it works.
- [ ] everything else should be fine, but running a few tests would be nice.

StrahilKazlachev:
fix #254


StrahilKazlachev:
@jods4 can you review?

jods4:
:shipit:

PR #291, opened 2017-05-08
feat(dialog-controller): deactivate hook error data

StrahilKazlachev:
Builds on top of #271 to provide consistent behavior to `DialogController.prototype.error` with the other *close* methods.
Main points of concern:
1. makes it easier to determine whether `DialogController.prototype.error` was called:
  - in the `.deactivate` hook
    - when `DialogController.prototype.error` is called, `.canDeactivate` is skipped.
    - `.deactivate` will be called with either `DialogCloseResult` or `DialogCloseError`(instance of `Error`)
  - in `dialogService.open({...}).whenClosed().catch()` - currently we can determine whether it is a cancellation error(`DialogCancelError`) or not(using just the error, without additional conventions, ...).
2. **breaking change** - currently `DialogController.prototype.error` accepts anything and rejects the result promise with that value. I propose that we wrap *whatever* is passed in a `DialogCloseError`, basically instance of `Error` with an additional field `reason: any` - the passed argument. It will break the current way of handling `dialogService.open({...}).whenClosed().catch()`, determining whether `DialogController.prototype.error` was called, since now the value/reason will be wrapped in a `DialogCloseError`.
  - this helps to determine  whether `.error` was called, by just checking the error param - used in **1.**

@PWKad @EisenbergEffect I would like to hear your concerns with **2.**
Will finish the tests afterwards.

### DIFF
--- a/src/dialog-controller.ts
+++ b/src/dialog-controller.ts
@@ -45,8 +45,8 @@ export class DialogController {
   /**
    * @internal
    */
-  public releaseResources(): Promise<void> {
-    return invokeLifecycle(this.controller.viewModel || {}, 'deactivate')
+  public releaseResources(dialogResult?: any): Promise<void> {
+    return invokeLifecycle(this.controller.viewModel || {}, 'deactivate', dialogResult)
       .then(() => this.renderer.hideDialog(this))
       .then(() => { this.controller.unbind(); });
   }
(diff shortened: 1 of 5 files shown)


EisenbergEffect:
I'm ok with this. We should definitely be rejecting with an error always anyways.

StrahilKazlachev:
Aligned the `DialogCloseError` with the rest of the *close* related API.
Currently we intentionally throw/reject only in 2 cases - when we cancel a dialog and `rejectOnCancel` is `true`, when `DialogController.prototype.error()` is invoked. Now we have dedicated errors for both cases.

StrahilKazlachev:
@PWKad could you review?

StrahilKazlachev:
@PWKad ping

src/dialog-controller.ts
```diff
@@ -97,25 +99,28 @@ export class DialogController {
...
-        return { wasCancelled: false };
-      }).catch(reason => {
+    const dialogResult = { wasCancelled: !ok, output } as DialogCloseResult;
+
+    return this.closePromise = invokeLifecycle(this.controller.viewModel || {}, 'canDeactivate', dialogResult)
+      .catch(reason => {
         this.closePromise = undefined;
         return Promise.reject(reason);
+      }).then(canDeactivate => {
+        if (!canDeactivate) {
```
PWKad: I think this chain of logic is already tested in the spec so should be fine, consumers of the API should see no difference in behavior correct?

StrahilKazlachev: Ah, yes. The only breaking change concerns `.error` - always rejects with `DialogCloseError` you can see my first 2 comments.

ydbondt:
Any idea when this feature is going to be released ?

EisenbergEffect:
I'm waiting to get confirmation from @StrahilKazlachev that we're ready for a release.

ydbondt:
Ok, thnx !

StrahilKazlachev:
@EisenbergEffect It's ready.

PR #302, opened 2017-06-16
Rc 1.0.0

PWKad:


StrahilKazlachev:
@PWKad I don't understand why are those additional *devDependencies* added. The dialog does not import anything from those new deps.

PWKad:
@StrahilKazlachev Tests were not running on a fresh clone.  I believe that aurelia-testing takes some additional peer dependencies as well that needed to be installed.

Here are a few for example -

https://github.com/aurelia/dialog/blob/master/test/unit/dialog-service.spec.ts#L1-L13

StrahilKazlachev:
@PWKad Ahhhhh, I see. Also it seems I've done quite a sloppy job in `test/main.ts` 😞. The interesting bit is that I don't get **any** errors, on windows, with clean checkout. That would also explain why the CI fails.
Lets merge and this week I'll try to find time to clean up any redundant dev/test deps, OK?

Issue #298, opened 2017-06-04
viewModel is not found by webpack


dkent600:
**I'm submitting a bug report**

* **Library Version:**
1.0.0-rc.1.0.3

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
6.10.3
* **NPM Version:**
5.0.0

* **JSPM OR Webpack AND Version**
webpack 2.5.1

* **Browser:**
All

* **Language:**
TypeScript 2.3

**Current behavior:**
DialogService is not finding my viewModel.  It does not report the problem and instead puts up an entirely empty and invisible dialog.

I invoke `.open`  like this:

```
        return this.dialogService.open(settings) ...
```

where `settings.viewModel` is set like this:
```
import {Prompt} from "../resources/commonDialogs/prompt";
.
.
.
settings.viewModel = Prompt;

```

(Note: `Prompt` is my own, not from the Aurelia-dialog example.)

DialogService sees  the `Prompt` viewModel function and tries to do something with it here:

```
 DialogService.prototype.ensureViewModel = function (compositionContext) {
        if (typeof compositionContext.viewModel === 'function') {
            compositionContext.viewModel = __WEBPACK_IMPORTED_MODULE_1_aurelia_metadata__["Origin"].get(compositionContext.viewModel).moduleId;
        }
        if (typeof compositionContext.viewModel === 'string') {
            return this.compositionEngine.ensureViewModel(compositionContext);
        }
        return Promise.resolve(compositionContext);
    };
```

But `WEBPACK_IMPORTED_MODULE_1_aurelia_metadata__["Origin"].get(compositionContext.viewModel).moduleId` returns `undefined`.

All of the relevant code is contained in a vendor.js file generated by webpack.

**Expected/desired behavior:**

* **What is the expected behavior?**
Should find my viewmodel or else report an error.  If there are additional steps that I'm not following, then they should be documented.

* **What is the motivation / use case for changing the behavior?**
Supplying a viewModel in the settings is supposed to work.  I assume webpack is part of the use case as well.

dkent600:
Would be nice to know if there are any workarounds for the time being.  At the moment I've got no dialogs...

StrahilKazlachev:
OK, finally found it - a nice [explanation](https://github.com/aurelia/dialog/issues/283#issuecomment-298885151) of what is happening. It is related to a separate issue but the `import` part is the same.

dkent600:
[...]

Issue #233, opened 2016-11-14
Dialog open in activate does not work on first load with router.


glennpierce:
**I'm submitting a bug report**

* **Library Version:**
1.0.0-beta.3.0.1

**Please tell us about your environment:**
* **Operating System:**
Linux fedora 23

* **Node Version:**
6.2.2

* **NPM Version:**
3.9.5

* **JSPM OR Webpack AND Version**
webpack 2.1.0-beta.22

* **Browser:**
Chrome 53

* **Language:**
TypeScript 2.0.0

**Current behavior:**

I am calling this.dialogService.open from an activate method.
When I first go to the page or refresh the page so the constructor is called the dialog open does not result in the dialog being displayed.
However, if I click on another entry in the top level menu provided by the router and then naviagate back to the original page the dialog displays fine. In ther later case the constructor is not called.
The only difference I can is is hasActiveDialog is true when the dialog displays but false when it does not work at the point open is called.

**Expected/desired behavior:**

Can't create a gistrun with aurelia-dialog so have made a repository at
https://github.com/glennpierce/test/tree/master/src

Go to any page and do a refresh and the dialog fails to display. Clinking between toplevel menu items will display the doalog.

* **What is the expected behavior?**

The dialog should display of first load.

* **What is the motivation / use case for changing the behavior?**

To display a dialog :)

StrahilKazlachev:
I'll have time to check it as early as tomorrow. If you want you can chain a `.catch` to the `.open` method and see the error, if any.

glennpierce:
Not sure if I'm doing anything wrong but I can't get a catch handler to trigger.

StrahilKazlachev:
@glennpierce It's strange that the catch doesn't work. Also I'm not much into webpack, but in the `webpack.config.js` I don't see anything about the dialog, is it normal? I'll first try it on the sample app first

StrahilKazlachev:
So I tried it with the skeleton-navigation/skeleton-esnext and it worked. When I try to run your sample I get `TypeError: webpack.validateSchema is not a function` after running `npm start`.

glennpierce:
I just used npm install aurelia-dialog --save like in the install instructions.
Don't think I needed to edit webpack.config.js

I have updated the source to solve the validateSchema error.

I had to use a different version of webpack

ie
-    "webpack": "2.1.0-beta.22",
-    "webpack-dev-server": ">=2.1.0-beta.0 || ^2.1.0"
[...]

PR #324, opened 2017-09-04
doc(dialog-basics): add webpack notes

StrahilKazlachev:
Added some notes related to #298.
@jods4 could you check if there is something I need to correct/rephrase/add?

### DIFF
--- a/doc/article/en-US/dialog-basics.md
+++ b/doc/article/en-US/dialog-basics.md
@@ -75,18 +75,21 @@ typings install github:aurelia/dialog --save
 
 <code-listing heading="main.js">
   <source-code lang="ES 2015">
+  import {PLATFORM} from 'aurelia-pal';
   export function configure(aurelia) {
     aurelia.use
       .standardConfiguration()
       .developmentLogging()
-      .plugin('aurelia-dialog');
+      .plugin(PLATFORM.moduleName('aurelia-dialog'));
 
     aurelia.start().then(a => a.setRoot());
   }
   </source-code>
 </code-listing>
 
-> Note: If you are using WebPack it is possible that the plugin is installed before Aurelia has replaced the `<body>` element, if that is where your `aurelia-app="main"` is defined, which results in some of the dialog components getting overwritten.  In this case you can move the `aurelia-app` attribute to a `<div>` inside of the `<body>`.  Example - `<body><div aurelia-app="main"></div></body>`.
+> Warning: When the `<body>` is marked with the `aurelia-app` attribute any dialog open prior to the app being attached to the DOM(before `Aurelia.prototype.setRoot` completes), will be *removed* from the DOM. Opening a dialog in the `canActivate` or `activate` hooks is *OK* in any scenario *if* you *await* it to close *before continuing*. If you just want to open a dialog, without awaiting it to close, do it the `attached` hook instead.
+
+> Warning: `PLATFORM.moduleName` should *not* be omitted if you are using Webpack.
 
 ## [Using The Plugin](aurelia-doc://section/3/version/1.0.0)
 
@@ -152,7 +155,6 @@ There is also an `output` property that gets returned with the outcome of the us
 <code-listing heading="edit-person.js">
   <source-code lang="ES 2015">
   import {DialogController} from 'aurelia-dialog';
-
   export class EditPerson {
     static inject = [DialogController];
     person = { firstName: '' };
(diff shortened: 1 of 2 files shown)


doc/article/en-US/dialog-basics.md
```diff
@@ -237,10 +241,28 @@ You can specify global settings as well for all dialogs to use when installing t
...
+    +-- app.js
+    +-- forms
+        +-- consent-form.js
+        +-- consent-form.html
+    +-- prompts
+        +-- prompt.js
+        +-- prompt.html
+    ```
+    If you want to open a `prompt` from `consent-form` the path will be `prompts/prompt`.
+    > Warning: WebPack users should use `PLATFORM.moduleName` when passing a path. Not using it may lead to the resource not being bundled, when not referenced anywhere else.
```
jods4: "when not referenced anywhere else"
It's more subtle than that. 
`moduleName` also ensures other properties important for `aurelia-loader`, like the assignment of a module id based on path, preserving exports and preventing their renaming, ensuring that the module doesn't get merged by `ModuleConcatenationPlugin`...

Bottom line: if a string is a dependency loaded by `aurelia-loader`, _always_ mark it as such with `moduleName`.

doc/article/en-US/dialog-basics.md
```diff
@@ -237,10 +241,28 @@ You can specify global settings as well for all dialogs to use when installing t
...
+  - object - it will be used as the view model. In this case `view` must also be specified.
+  - class - the view model *class* or *constructor function*.
+    > Note: This approach depends on being able resolve the origin of the class.
+
+    > Warning: Known WebPack issues when using *DllPlugin*, *ModuleConcatenationPlugin*.
```
jods4: I would rephrase: "this approach might not work when using `DLLPlugin` or `ModuleConcatenationPlugin`."
And add a note to the effect of:
"Webpack users are advised to always use the `string` overload wrapped with `moduleName`"

jods4:
I proposed tiny precisions to your changed. Other than that, looked fine to me!

StrahilKazlachev:
@jods4 I didn't see any info on `PLATFORM.moduleName` in the Webpack section of the hub and why it is **important**. So I hope you don't mind I took one of your comments as is.

jods4:
The hub still describes the v1.x plugins.
In the meantime, there's a lot of info about the new v2.x plugins in `aurelia-webpack-plugin` wiki, though.

Webpack traces a precise graph of each module's dependencies. 
`moduleName` is important because it teaches webpack about `aurelia-loader` dynamic dependencies.

If your project is not with Webpack you don't have to use it.
If you build a library or share code with others, then you should. It doesn't hurt to have it in the code as `aurelia-pal` implementation is just a no-op.

StrahilKazlachev:
@jods4 If you are OK with the current changes I'll merge it.
About `PLATFORM.moduleName`, I just wanted to point to more info about it(or add in the dialog docs, as I did), so that we're not just saying that it *must be used* but and *why*.

doc/article/en-US/dialog-basics.md
```diff
@@ -75,18 +75,21 @@ typings install github:aurelia/dialog --save
...
-      .plugin('aurelia-dialog');
+      .plugin(PLATFORM.moduleName('aurelia-dialog'));
 
     aurelia.start().then(a => a.setRoot());
   }
   </source-code>
 </code-listing>
 
-> Note: If you are using WebPack it is possible that the plugin is installed before Aurelia has replaced the `<body>` element, if that is where your `aurelia-app="main"` is defined, which results in some of the dialog components getting overwritten.  In this case you can move the `aurelia-app` attribute to a `<div>` inside of the `<body>`.  Example - `<body><div aurelia-app="main"></div></body>`.
+> Note: If you are using Webpack it is possible that the plugin is installed before Aurelia has replaced the `<body>` element, if that is where your `aurelia-app="main"` is defined, which results in some of the dialog components getting overwritten.  In this case you can move the `aurelia-app` attribute to a `<div>` inside of the `<body>`.  Example - `<body><div aurelia-app="main"></div></body>`.
```
jods4: Reading this again and it looks nasty to me.
Isn't that a timing bug in `aurelia-dialog`?
If you put stuff inside the `<body>` couldn't you just wait on Aurelia's loaded event?

jods4:
Looks ok to me, but I just noted that the first addition is quite nasty. 
It's the kind of things that should "just" work out of the box :(

As far as explaining _why_ the best I can offer is to point users to this documentation: https://github.com/aurelia/webpack-plugin/wiki

StrahilKazlachev:
@jods4 You are right about the nasty part. The reason I mentioned the Hub was I wanted to link to it, not the wiki, but it'll be probably better.
Thanks for the help 😄

StrahilKazlachev:
@jods4 So I tracked why that *nasty* part was added - https://github.com/aurelia/dialog/issues/233#issuecomment-260755619, I'll see how to *properly* address it.

jods4:
If I might: I don't like advice without explanation as it tends to turn into cargo cult.

Instead of just saying that it might get dismissed, could you say under which conditions? "when your app is the <body> tag, any dialog opened by the first page loaded in the application before `attached` and not properly awaited will be dismissed"

jods4:
'the body tag

PR #326, opened 2017-10-15
feat(dialog-renderer): add Animator support close #188

### LINKED ISSUE #188: FR: Integrate with animator
- **Library Version:**
  1.0.0-beta.1.1.0

- **Operating System:**
  Linux (Ubuntu 16.04)
- **Node Version:**
  6.2.2
- **NPM Version:**
  3.9.5
- **JSPM OR Webpack AND Version**
  JSPM 0.16.29
- **Browser:**
  Chrome 51
- **Language:**
  TypeScript 1.8.9

**Current behavior:**
If one or more transitions are defined on the dialog container, then the renderer will [wait for the first one to finish](https://github.com/aurelia/dialog/blob/1.0.0-beta.2.0.0/src/dialog-renderer.js#L117) before showing a dialog. Conversely, when the last dialog is closed, the renderer will [wait for the first transition to finish](https://github.com/aurelia/dialog/blob/1.0.0-beta.2.0.0/src/dialog-renderer.js#L154) on the dialog container before removing the dialog overlay and container from the DOM.

**What is the expected behavior?**
Ideally, all animation for the opening and closing of dialogs should be up to the user (with a nice looking default perhaps), and the renderer should await all animations that occur during the opening or closing of dialogs.

In the case of the `CssAnimator` for example, one might expect the renderer to wait for the closing animation before removing the dialog and dialog container from the DOM:

``` html
<template>
  <ai-dialog class="au-animate">
    ...
  </ai-dialog>
</template>
```

``` css
@keyframes fadeOut {
    0% { opacity: 1; }
    100% { opacity: 0; }
}
ai-dialog.au-animate.leave-active {
  animation: fadeOut 2s ease-in-out;
}
```

**What is the motivation / use case for changing the behavior?**
Awaiting animations after manipulating nodes is already [standard behavior in templating](https://github.com/aurelia/templating/blob/1.0.0-rc.1.0.1/src/view-slot.js#L299).

StrahilKazlachev:
1. Extracted the composition logic from `DialogService` in a new `DialogCompositionEngine` - was getting messy.
2. `DialogController`
    1. Marked the `.controller` property as *internal* - **can be** a **breaking change**(TS only) if someone used it outside of a `Renderer` implementation, which I think should not be done.
    2. Added `.view` internal property.
    3. Introduced new `interface InfrastructureDialogController extends DialogController`
        - reintroduces the `.controller` and `.view` properties - makes them non *internal*.
        - I really think these properties *should* only be used by `Renderer` implementations.
        - a *better* name is welcomed.
3. `Renderer` - switched `DialogController` with `InfrastructureDialogController`.
4. `DialogRenderer`
    1. integrated with the `Animator`
    2. removed the current transition based animation support - **breaking change** - *I* don't see how to support both.

TODO:
- [ ] deprecate the `ignoreTransitions` setting - update docs, leave in interface for now
- [ ] update docs regarding the `Animator` support
- [ ] update the tests


EisenbergEffect:
Ultimately, I always wanted this to use the Animator interface. Glad to see this work is progressing!

StrahilKazlachev:
1. I think that we should let the `Renderer` decide whether to use a `ViewSlot` or something else, we should just provide it with the bound `Controller`/`View` of the dialog *content*(the user defined piece).
    - pass a *noop* `ViewSlot` to `CompositionEngine.prototype.compose` - this way the `View` will not be added to anything prior to providing it to the `Renderer`. Didn't come up with other solution how to prevent `.compose` from adding the `View` to a `ViewSlot`.
2. For now we should probably let `Renderer`s decide what of their layout should be animated and how.
    - for the `DialogRenderer` I've animated the `ux-dialog-overlay` if `ViewSlot.prototype.add`/`ViewSlot.prototype.remove` return a `Promise`.

EisenbergEffect:
@PWKad Can you help by reviewing this PR?
@StrahilKazlachev Can you provide some more context around the noop ViewSlot bit?

StrahilKazlachev:
@EisenbergEffect About the *noop* `ViewSlot`.
The main reason is I *only* want the `CompositionEngine` to create/compose a `Controller`/`View`, I don't want it to add them(the controller view or just the view) to a `ViewSlot`. What I gain:
1. A `Renderer` can now create the `ViewSlot` itself, no one else actually needs it.
    - can instantiate it with specific `Animator`.
2. A `Renderer` can now **first** *attach* the `ViewSlot` and **then** call `ViewSlot.prototype.add`.
    - the `View` will be *attached*, then if is marked for animation it will it will be *animated* - all this will be done by the `ViewSlot`, a `Renderer` only needs to properly handle the result of `ViewSlot.prototype.add` - I'm aware of `ViewSlot.prototype.animateView`.
    - when `Renderer.prototype.hideDialog` is called the approach will be *symmetrical*, just call `ViewSlot.prototype.remove` and handle the result appropriately.
3. A more sophisticated `Renderer` could add the `Controller`/`View` to its own `View` hierarchy - still exploring this scenario.

It's not like we can't:
1. create the `ViewSlot` outside of the `Renderer`.
2. expose it through the `DialogController`.
3. after the `CompositionEngine.prototype.compose` completes call `ViewSlot.prototype.remove` or every `Renderer` can do this if it decides to.
4. proceed calling `Renderer.prototype.showDialog`.

But it but it doesn't feel right to me.

EisenbergEffect:
Gotcha. Thanks for the explanation.

test/unit/dialog-renderer.spec.ts
```diff
@@ -270,97 +254,99 @@ describe('DialogRenderer', () => {
...
-        await show(done, renderer);
-        spyOn(renderer.dialogContainer, 'removeEventListener').and.callThrough();
-        await hide(done, renderer);
-        expect(renderer.dialogContainer.removeEventListener)
-          .toHaveBeenCalledWith(transitionEvent(), jasmine.any(Function));
-        done();
-      });
-    });
-  });
+  // TODO: substitue with Animator integration tests
```
PWKad: Are these still a work in progress?

src/dialog-controller.ts
```diff
@@ -102,7 +162,9 @@ export class DialogController {
 
     const dialogResult: DialogCloseResult = { wasCancelled: !ok, output };
 
-    return this.closePromise = invokeLifecycle(this.controller.viewModel || {}, 'canDeactivate', dialogResult)
+    // tslint:disable-next-line:max-line-length
+    // TODO: add canDeactivate API
```
PWKad: Still in process?

src/dialog-controller.ts
```diff
@@ -33,24 +53,64 @@ export class DialogController {
...
+            this.reject = reject;
+          })
+        };
+      });
+  }
+
+  // /**
+  //  * @internal
+  //  */
+  // public hide(): Promise<void> {
```
PWKad: commented code

PWKad:
Awesome work @StrahilKazlachev 

Looks like there is still some work in process items, let me know once it is ready and I can pull it down and test further.

src/dialog-controller.ts
StrahilKazlachev: Leftover, I decided not to do it in this PR.

src/dialog-controller.ts
StrahilKazlachev: Leftover, I decided not to do it in this PR.

test/unit/dialog-renderer.spec.ts
StrahilKazlachev: Still working on the tests.

StrahilKazlachev:
As a whole only the tests and docs are WIP,  I may refactor some more code, but I'm *done* with the API changes - depends on feedback.

EisenbergEffect:
How are we doing on this? Ready to merge?

jeffgrann:
This issue has been around for a long time. Is this fix going to be in the next release?

StrahilKazlachev:
@PWKad Quite late, but lets say I'm done for now.
One thing I can't really decide on - having 2 `DialogController` interfaces:
- `DialogController` - min API surface relevant to app code.
- `InfrastructureDialogController extends DialogController` - adds additional APIs relevant to `Renderer` implementations - `view: View`, `controller?: Controller`. May be expanded with new features for the renderers.

In my code I would go with having both *interfaces*, but here I'm not so sure.

EisenbergEffect:
@StrahilKazlachev I think the two interfaces is fine. Are you ready for final review and merge?

src/dialog-composition-engine.ts
```diff
@@ -0,0 +1,137 @@
...
+            model: controller.settings.model,
+            view: controller.settings.view,
+            viewModel: controller.settings.viewModel,
+            // we don't want the CompositionEngine to add the View to a ViewSlot,
+            // that's the job of the Renderer
+            viewSlot: noopViewSlot,
+            host: controller.renderer.getDialogContainer()
+        };
+        if (typeof context.viewModel === 'function') {
+            const moduleId = Origin.get(context.viewModel).moduleId;
```
EisenbergEffect: Isn't this needed only if we don't have a view?

StrahilKazlachev: @EisenbergEffect no, this is the case where you pass a `class` for the `viewModel`.
```
import {MyDialogViewModel} from './my-dialog';
//...
dialogService.open({viewModel: MyDialogViewModel})
//...
```
And this is the case that has the *greatest* issues with `WebPack` and needs a bit more work from the one using it to make it work.
This is supported by the dialog from the early `beta`s.

StrahilKazlachev:
@EisenbergEffect No documentation update in this PR. Was thinking of adding it later. Will try this week.
Ignoring the lack of docs update, it is ready for final review.

src/dialog-composition-engine.ts
EisenbergEffect: Cool. Just checking. It's been a while since I looked at this ;)

EisenbergEffect:
@StrahilKazlachev Can you work on reviewing the docs and seeing if any updates are needed? I can push out a dialog and doc release simultaneously. Just let me know.

PR #332, opened 2017-11-11
Fixed the markdown for code snippets

kabaehr:
... and added install command for Aurelia CLI

Presumably, the text was copied from Aurelia Hub or somewhere else and got not adapted for github


StrahilKazlachev:
The CLI install command is OK, but the other changes will break the docs site. The format may not be github friendly, but for now this is it.

EisenbergEffect:
Don't want to change the format right now. It would make the aurelia site rendering worse. We hope to make some improvements to the site so that more basic markdown can be used for code in the future.

Issue #335, opened 2018-01-12
Should be based on web standards HTML <dialog> element


timfish:
Aurelia is forward looking and supports web standards so shouldn't aurelia-dialog be built using the [HTML dialog element](https://www.w3.org/TR/html52/interactive-elements.html#the-dialog-element), be based around that API and use the relevant polyfils to aid with older browsers?

Obviously the API is not as nice as the `aurelia-dialog` API (ie. no async) so it might be worth wrapping it but in Chrome 64 testing [these examples](https://demo.agektmr.com/dialog/) `showModal()` appears to fix #1 (focus issues) and will no doubt be more accessible to screen readers than the current implementation.

timfish:
The HTML 5.2 specification [mentions an `autofocus` attribute](https://www.w3.org/TR/html52/sec-forms.html#autofocusing-a-form-control-the-autofocus-attribute) which works along with the dialog spec to set the focus in the dialog. This would replace the aurelia-dialog `attach-focus` which does the same thing.

zewa666:
Great idea, now lets hope all the browsers and electron catch up soon enough

timfish:
It was added to chrome 37 so we don't have to worry about electron and the polyfill seems to work well as it appears to fix #1 on Safari.

StrahilKazlachev:
So lets see if this can be cleanly added in a `Renderer`.

RomkeVdMeulen:
As for accessibility: in my experience most screenreaders are still somewhat lagging in HTML5 support, but hopefully they'll catch up over time. I don't know how well the `<dialog>` is / can be made backwards compatible using ARIA. I'll do some testing and come up with some recommendations.

timfish:
For backwards compatibility I think we'd probably have to include:
```html
<dialog role="dialog" aria-modal="true">
</dialog>
```
`aria-modal` was introduced in ARIA 1.1. To support ARIA 1.0 screenreaders, anything outside the dialog should have `aria-hidden="true"` added which should have a similar effect.

RomkeVdMeulen:
Some quick testing tells me support for the `<dialog>` tag is surprisingly good on Windows. Chrome seems to support the tag natively: when using NVDA 2017.4 and Chrome 63, the dialog is announced. NVDA with IE 11 or Firefox doesn't have native support, but the polyfill already sets `role="dialog"` automatically and with that, the dialog is announced on NVDA as well.

[...]

PR #338, opened 2018-01-22
Add alternative <dialog> element renderer

### LINKED ISSUE #1: focus considerations
(moved from https://github.com/aurelia/framework/issues/142)
1. [`.autofocus`](https://github.com/BlueSpire/Durandal/blob/master/src/plugins/js/dialog.js#L502) / [`autofocus`](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Forms_in_HTML#The_autofocus_attribute): 
2. [preventing users from tabbing off modals onto the underlying page](https://github.com/angular-ui/bootstrap/issues/738#issuecomment-24702467) 
3. Ensuring the focus is restored to the previously focused element when the modal is closed.
4. Nested modals: Are all the scenarios above handled when multiple modals are opened?

timfish:
I wont repeat [everything I've already said in #335](https://github.com/aurelia/dialog/issues/335#issuecomment-359568900). 

This would finally close #1 and improve accessabilty.

**Breaking Changes**
- ~~`startingZIndex` setting is no longer required~~
- `position` callback is no longer passed overlay element as it no longer exists
- `centerHorizontalOnly` is not used. CSS only.

**To do**
- [x] Test that everything all works in polyfilled browsers
  - [x] Karma tests pass in Firefox 58.0 (2 transition tests still failing)
  - [x] Test real world usage in some other browsers

### DIFF
--- a/src/aurelia-dialog.ts
+++ b/src/aurelia-dialog.ts
@@ -27,3 +27,4 @@ export * from './dialog-cancel-error';
 export * from './dialog-result';
 export * from './dialog-service';
 export * from './dialog-controller';
+export { NativeDialogRenderer } from './native-dialog-renderer';
--- a/src/dialog-settings.ts
+++ b/src/dialog-settings.ts
@@ -82,7 +82,7 @@ export interface DialogSettings {
    * Usde to provide custom positioning logic.
    * When invoked the function is passed the dialog container and the dialog overlay elements.
    */
-  position?: (dialogContainer: Element, dialogOverlay: Element) => void;
+  position?: (dialogContainer: Element, dialogOverlay?: Element) => void;
 }
 
 /**
--- a/src/native-dialog-renderer.ts
+++ b/src/native-dialog-renderer.ts
@@ -1,0 +1,187 @@
+import { DOM } from 'aurelia-pal';
+import { transient } from 'aurelia-dependency-injection';
+import { Renderer } from './renderer';
+import { DialogController } from './dialog-controller';
+import { transitionEvent, hasTransition } from './dialog-renderer';
+
+const containerTagName = 'dialog';
+let body: HTMLBodyElement;
+
+@transient()
+export class NativeDialogRenderer implements Renderer {
+  public static dialogControllers: DialogController[] = [];
+
+  public static keyboardEventHandler(e: KeyboardEvent) {
+    const key = (e.code || e.key) === 'Enter' || e.keyCode === 13
+    ? 'Enter'
+    : undefined;
+
+    if (!key) { return; }
+    const top = NativeDialogRenderer.dialogControllers[NativeDialogRenderer.dialogControllers.length - 1];
+    if (!top || !top.settings.keyboard) { return; }
+    const keyboard = top.settings.keyboard;
+    if (key === 'Enter' && (keyboard === key || (Array.isArray(keyboard) && keyboard.indexOf(key) > -1))) {
+      top.ok();
+    }
+  }
+
+  public static trackController(dialogController: DialogController): void {
+    if (!NativeDialogRenderer.dialogControllers.length) {
+      DOM.addEventListener('keyup', NativeDialogRenderer.keyboardEventHandler, false);
+    }
+    NativeDialogRenderer.dialogControllers.push(dialogController);
+  }
+
+  public static untrackController(dialogController: DialogController): void {
+    const i = NativeDialogRenderer.dialogControllers.indexOf(dialogController);
+    if (i !== -1) {
+      NativeDialogRenderer.dialogControllers.splice(i, 1);
+    }
+    if (!NativeDialogRenderer.dialogControllers.length) {
+      DOM.removeEventListener('keyup', NativeDialogRenderer.keyboardEventHandler, false);
+    }
+  }
+
+  private stopPropagation: (e: MouseEvent & { _aureliaDialogHostClicked: boolean }) => void;
+  private closeDialogClick: (e: MouseEvent & { _aureliaDialogHostClicked: boolean }) => void;
+  private dialogCancel: (e: Event) => void;
+
+  public dialogContainer: HTMLDialogElement;
+  public host: Element;
+  public anchor: Element;
+
+  private getOwnElements(parent: Element, selector: string): Element[] {
+    const elements = parent.querySelectorAll(selector);
+    const own: Element[] = [];
+    for (let i = 0; i < elements.length; i++) {
+      if (elements[i].parentElement === parent) {
+        own.push(elements[i]);
+      }
+    }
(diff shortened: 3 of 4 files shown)


timfish:
Looks like transitions are broken by the Chrome default CSS:
```css
dialog:not([open]) {
  display: none;
}
```

timfish:
The `transition*` events don't appear to fire when the dialog is opened via `showModal()`. This might have something to do with it being moved to the "top layer"?

When manually adding and removing the `open` attribute the transition events appear to work. We can't however manually add/remove the `open` attribute as we would lose all the dialog goodness.

StrahilKazlachev:
@timfish Ignore the transitions part, it will be gone soon.

timfish:
The tests are still failing but I've managed to get transitions working in normal usage by having transition on `.active` rather than the `open` attribute and only calling `<dialog>.close()` after the transition is over. 

@StrahilKazlachev do you know why `centerHorizontalOnly` is even required? Can't all positioning be via CSS or `position` callback?

timfish:
All but 4 of the transition tests now pass in Firefox with the polyfill.

timfish:
I've now tested this working on Chrome, Chrome Android, Firefox, Safari Mac, Safari iOS 11, Edge and IE11.

They only differ on dialog positioning and checking the [browser support](https://caniuse.com/#search=fit-content) this is entirely down to the use of `fit-content`:

**Chrome, Chrome Android, Safari Mac, Safari iOS 11**
Vertically and horizontally centred 

**Firefox**
Horizontally centred at the top

**Edge and IE11**
Top left

Can we just go flex and remove `centerHorizontalOnly`?

StrahilKazlachev:
I took my time to think on this and for now I don't like the `<dialog>` element. It feels incomplete, and troublesome.
Still, since the community has interest in it I propose we add one more *out-of-the-box* `Renderer` - `DialogRendererNative`. 
Having 2 *Aurelia* renderers will also give us more insight on how to improve our `Renderer` and related APIs. 
@EisenbergEffect Your thoughts?

timfish:
@StrahilKazlachev That's a good idea, I'm already using this renderer like:
```
    .plugin('aurelia-dialog', (config: DialogConfiguration) => {
      config.useDefaults();
      config.useRenderer(CustomDialogRenderer)
    })
```

EisenbergEffect:
Go for it! 😊

timfish:
@StrahilKazlachev I'll rework this into its own renderer and revert all the other changes.

dannyBies:
Is this ready to be merged?

timfish:
It doesn't touch any existing code so it's pretty safe to merge.

The docs will need some updating to show how to use the alternative renderer but it's probably worth it getting some testing in a beta first.

jmezach:
@timfish I think @dannyBies and I would be more than happy to beta test this ;).

StrahilKazlachev:
This is not a separate package - the new `Renderer`, so can't do beta since we are currently in RC*ish*.
What is still bugging me is how **not** to export the `Renderer`s and still be able to set them. For now we don't want them to be used as base classes.

timfish:
Sounds like the easiest thing to do for now would be to just host this renderer in another package? It just makes maintenance a bit of a pain.

For those wanting to test it now, if you're using TypeScript you can just copy `dialog-renderer-native.ts` into your project, fix the imports and use it like: 
```
    .plugin('aurelia-dialog', (config: DialogConfiguration) => {
      config.useDefaults();
      config.useRenderer(DialogRendererNative)
    })
``` 

For JavaScript you'll have to convert it or build and use the dist output.

StrahilKazlachev:
OK, so lets make this to 1 commit with the appropriate commit message and I'll get it in.

timfish:
Do you not just squash it into a single commit on merge?

Alexander-Taran:
@timfish it does.. but commit message - i don't remember

CasiOo:
Looks like a lot of code is copy-pasted from `dialog-renderer.ts`. Can't we just use the exported functions from `dialog-renderer.ts` instead of copy-pasting code around :)?

Rename `DialogRendererNative` to `NativeDialogRenderer`

Alexander-Taran:
@timfish I'd suggest then if you don't want to rebase / squash yourself just to provide a "final" commit message so it'd be easier to copy/paste it when merge/squash will happen (-:

timfish:
@Alexander-Taran I'll rebase it.

Looking through the duplication between the two renderers I could reduce it by making `NativeDialogRenderer` extend `DialogRenderer`. This might complicate things as they further diverge?

StrahilKazlachev:
@timfish correct about the complications, there is a lot of work to be done around the `Renderer`s in my opinion. That's why I'm pleased to have a second *out-of-the-box* impl, even if currently it complicates some aspects.
Lets leave it as is.

timfish:
I did try rebasing but it seems beyond me this morning. If you can squash merge this with a commit message something like below, that would be great:

```
feat(dialog-renderer): Add native dialog renderer
```

timfish:
Perhaps the docs could do with some details on using the new renderer?

```
import { NativeDialogRenderer } from 'aurelia-dialog';

aurelia.use.plugin('aurelia-dialog', (config: DialogConfiguration) => {
  config.useDefaults();
  config.useRenderer(NativeDialogRenderer)
})
```

It will also require custom CSS to stop it looking awful.

StrahilKazlachev:
Will do later today.
Actually the `Renderer` css is what is bugging me. Separating the current inlined css will be easy for the custom elements, but for the renderers I have no good idea.

timfish:
Yeah, the CSS should probably be renderer specific and editing the inline string is fragile at best!

Ah, also worth noting that native dialog needs [polyfilling](https://github.com/GoogleChrome/dialog-polyfill) still on many browsers.

Alexander-Taran:
I'd leave styling of native to the end user.. so it would require some fiddling and discovering the need in polifil as well (-:

rdemetrescu:
Wow, amazing work! I really would like to see it merged soon.

dannyBies:
It would be really great to see this merged!

RomkeVdMeulen:
@timfish If I understand correctly, to polyfill this I need to do something like this:

```js
window.dialogPolyfill = await SystemJS.import("dialog-polyfill"); // or require() or import() or whatever
```

Did I get that right?

timfish:
Its expecting `window.dialogPolyfill` to be set and the easiest way to achieve this is to just load `dialog-polyfill.min.js` into a script tag in your html.

glen-84:
No documentation?

Alexander-Taran:
@glen-84  there.. I created a pull request (-:
https://github.com/aurelia/documentation/pull/486

PR #343, opened 2018-03-14
Trucante long text in display table

victorDivino:
The table should not grow outside the width of the screen, if displays text (potentially long), which must be truncated when necessary, so that the text fits into the child elements, which is adapted to the width of the table, which may not be larger than the width of the screen.

[bad](https://codepen.io/victorDivino/pen/MVyrag)
[fix](https://codepen.io/victorDivino/pen/gerowj)

### DIFF
--- a/styles/input.less
+++ b/styles/input.less
@@ -55,6 +55,7 @@ ai-dialog-container {
 ai-dialog {
   width: 100%;
   display: table;
+  table-layout: fixed;
   box-shadow: 0 5px 15px rgba(0,0,0,.5);
   border: 1px solid rgba(0,0,0,.2);
   border-radius: 5px;


Alexander-Taran:
Could you attach 2 screenshots for a bad rendering before and after fix?

victorDivino:
[Eg not working](https://codepen.io/victorDivino/pen/MVyrag)
[Eg working](https://codepen.io/victorDivino/pen/gerowj)

victorDivino:
# Without `table-layout: fixed`
pt1

# With `table-layout: fixed`
pt2

As you can see, the modal width does't grow outside the screen when we use fixed layout!

**Emails were blured for security purpose**

Alexander-Taran:
Not sure about this.
@StrahilKazlachev what do you think?

AlbertoMonteiro:
@Alexander-Taran take look here: https://css-tricks.com/fixing-tables-long-strings/
You can see in first two samples, that when table does not have the layout fixed, the width of table, can be higher of the table's parent element, in the second sample, using fixed layout the last 3 tables that in previous sample had overflowed, doest do that anymore.

Alexander-Taran:
Yeah Chris Coyier is the man (-:
@StrahilKazlachev & @EisenbergEffect 
pretty safe to merge

unless we have sevaral 'rows' for the table that is dialog
otherwise width will be calculated by 1st row

EisenbergEffect:
Looks good to me. @StrahilKazlachev Can we get this in? Any objections?

JSeligsohn:
Any reason this wasn't merged? I just discovered this fix on my own and would be nice if had been merged already to save me (and others) from a headache :)

Issue #344, opened 2018-03-16
Dialog VM hooks for keyboard (Enter/Escape)


pape87:
**I'm submitting a feature request**

**Current behavior:**
When settings.keyboard is configured to respect keyboard events, upon pressing the key (Enter or Escape), only controller.ok()/controller.cancel() are called respectively.
There is no opportunity to verify or pass a result.

**Expected/desired behavior:**
Introduce two new life-cycle hooks to the dialog VM like this:

```typescript
import { DialogController } from "aurelia-dialog";

export class DialogVM {
  constructor(private ctrl: DialogController) {
  }

  public ok() {
    // result of this function is passed to controller.ok()
    // if a rejected promise is returned or an error thrown do not close the dialog
  }

  public cancel() {
    // result of this function is passed to controller.cancel()
    // if a rejected promise is returned or an error thrown do not close the dialog
  }
}

```

StrahilKazlachev:
1. Yes, there is no way to pass custom data.
2. I don't understand *verify*. To verify what? When calling `.ok()`/`.cancel()` if the VM implements `.canDeactivate`, that hook will be called with a `DialogCloseResult`. So you can see what *intention* it is being closed and prevent the closing if needed by returning `false`.
3. The keyboard logic is all in the *default* `Renderer` implementation. And yes some `Renderer` options are mixed in the general ones, this has been since day one.

Alexander-Taran:
Oh my!!
and it is referenced in the docs
http://aurelia.io/docs/plugins/dialog#lifecycle-hooks

never knew

StrahilKazlachev:
Now that I think about it, both `canDeactivate` and `deactivate` hooks are called with `DialogCloseResult`. So you can manipulate the `.output` field - definitely not intended and feels like a hack but it is doable, since we are not freezing the result object(don't think we'll do it).

Alexander-Taran:
@pape87 do you think this satisfies you?

pape87:
Yes this solved my issue. I've added some info to the docs about the canDeactivate result parameter.

Thank you

Alexander-Taran:
@pape87 take a look at this comment about interfaces available for dialog vm
https://github.com/aurelia/framework/issues/874#issuecomment-374203146

Alexander-Taran:
@pape87 thanks for your help with docs (-:

PR #345, opened 2018-03-19
docs(hooks): canDeactivate result parameter

pape87:
Add information about the result parameter

Fixes issue: https://github.com/aurelia/dialog/issues/344

### DIFF
--- a/doc/article/en-US/dialog-basics.md
+++ b/doc/article/en-US/dialog-basics.md
@@ -358,9 +358,10 @@ With this hook you can cancel the opening of a dialog. It is invoked with one pa
 
 This hook can be used to do any necessary init work. The hook is invoked with one parameter - the value of the `model` setting passed to `.open()`.
 
-### `.canDeactivate()`
+### `.canDeactivate(result: DialogCloseResult)`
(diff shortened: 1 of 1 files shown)


Alexander-Taran:
I'm afraid that won't cut it.
If I remember right deactivate() also receives dialogResult.
And you might want to annotate the code with typing info. 
Last one I'm not sure of. 
Because there is no inheritance.

zewa666:
Right so deactivate takes `result: DialogCloseResult | DialogCloseError`
https://github.com/aurelia/dialog/blob/458477cf7ad3a517280bd97efb62045aa5ea5bba/src/interfaces.ts

This should be updated in the docs PR as well, but I wouldn't annotate any code or so as this is solely a docs PR, so that people reading the documentation immediately figure out there is something passed into the lifecycle hooks.

Alexander-Taran:
@zewa666 agreed. There are interfaces already in place.
https://github.com/aurelia/framework/issues/874#issuecomment-374226001

Alexander-Taran:
PR looks good.
can be merged

PR #347, opened 2018-05-31
chore(all): update aurelia-binding dependency

fkleuver:
@EisenbergEffect  @bigopon


StrahilKazlachev:
@EisenbergEffect  @fkleuver was the `package-lock.json` added intentionally?

fkleuver:
Actually no .. not sure how that one slipped past. Will delete it again. Thanks for the notice!

PR #349, opened 2018-06-01
fix(attach-focus): restore behavior prior to #346

StrahilKazlachev:
make it work as boolean attribute
set the default binding mode to oneTime
update tests

### DIFF
--- a/test/unit/attach-focus.spec.ts
+++ b/test/unit/attach-focus.spec.ts
@@ -86,10 +86,9 @@ describe('attach-focus', () => {
     component.create(bootstrap).then(() => {
       expect(document.activeElement).not.toBe(viewModel.focusTargetElement);
       viewModel.hasFocus = true;
-      (Container.instance.get(TaskQueue) as TaskQueue).queueMicroTask(() => {
-        expect(document.activeElement).not.toBe(viewModel.focusTargetElement);
-        done();
-      });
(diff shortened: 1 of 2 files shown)


StrahilKazlachev:
@bigopon can you check? As we discussed, I hope so 😅

test/unit/attach-focus.spec.ts
```diff
@@ -18,50 +30,69 @@ describe('modal gets focused when attached', () => {
...
     });
   });
 
+  it('sets focus only in ".attached"', done => {
+    setupView(component, 'attach-focus.to-view="hasFocus"');
+    viewModel.hasFocus = false;
+    component.create(bootstrap).then(() => {
+      expect(document.activeElement).not.toBe(viewModel.focusTargetElement);
+      viewModel.hasFocus = true;
+      (Container.instance.get(TaskQueue) as TaskQueue).queueMicroTask(() => {
```
StrahilKazlachev: @bigopon can you advice if there is a better way to await the change?

bigopon: task queue has `flushMicroTaskQueue()` that can be used for this situation, to immediately apply the changes

StrahilKazlachev: @bigopon I'll still need the `TaskQueue`. Anything in *testing* I can use?

PR #350, opened 2018-06-02
feat(all): support non-modal dialogs

StrahilKazlachev:
I'm proposing to support non-modal dialogs - the native dialogs support it.
1. A **modal dialog** is always *on top*(visually)
    - `Renderer`s will have to append any **non-modal dialog** before existing **modal** ones in the *host*
    - the `DialogService` `.controllers` property will still keep the controllers in the order they were open
2. Introduce a *focused* concept - the dialog the user is currently interacting with(*accepting naming propositions*)
    - apply actions on the *focused* dialog
        - whether or not and on which dialog, to apply the **keyboard** actions when there are only **non-modal dialogs**
    - the top **modal dialog** is always the *focused*
    - when there are no **modal dialogs** open there **may** or may **not** be a *focused* dialog

Did just minor changes to start a discussion.
@EisenbergEffect @PWKad what do you think?

### DIFF
--- a/src/dialog-renderer-default.ts
+++ b/src/dialog-renderer-default.ts
@@ -73,7 +73,12 @@ export class DialogRendererDefault implements Renderer {
   }
 
   private setupClickDismissHandling(dialogController: InfrastructureDialogController): void {
-    this.contentInboundClick = eo => { eo[CONTENT_INBOUND_CLICK] = true; };
+    this.contentInboundClick = eo => {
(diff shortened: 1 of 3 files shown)


bigopon:
Would be nice if we could expose the `focus` state on dialog controller, so that non blocking dialogs can leverage it to set proper z index

davismj:
http://davismj.me/blog/easy-dialog/

PR #351, opened 2018-06-18
Use fixed versions in typings.json

fracz:
Using the always the latest version leads to unstable builds. Currently, all projects that use TS and the `aurelia-dialog` fails to build because of [the breaking change in the `aurelia-dependency-inection` v1.4.0](https://github.com/aurelia/dependency-injection/issues/166#issuecomment-397985431).

### DIFF
--- a/typings.json
+++ b/typings.json
@@ -2,9 +2,9 @@
   "name": "aurelia-dialog",
   "main": "dist/commonjs/aurelia-dialog.d.ts",
   "dependencies": {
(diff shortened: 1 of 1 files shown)


StrahilKazlachev:
The lack of response was not intentional. Did think on this and for me this is substituting one problem with another. It is just as possible that this PRs behaviour will become the unwanted case - locking down the deps versions. So that's my reason for closing.

PR #363, opened 2018-12-12
feat(keyboard): allow to configure keyevent type

zewa666:
when configuring the plugin one can pass the keyEvent as a setting, which accepts 'keyup' and 'keydown'

### DIFF
--- a/src/dialog-keyboard-service.ts
+++ b/src/dialog-keyboard-service.ts
@@ -1,6 +1,6 @@
 import { DOM } from 'aurelia-pal';
 import { DialogController } from './dialog-controller';
-import { ActionKey } from './dialog-settings';
+import { ActionKey, KeyEventType } from './dialog-settings';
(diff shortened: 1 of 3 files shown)


zewa666:
@StrahilKazlachev this was made because of a very funny issue I had. In our app we're using https://github.com/bevacqua/horsey to simulate a command palette, such as the F1 palette in VSCode, to offer various features for easier discovery. Now one of the selectable options will open up an Aurelia dialog. And funny enough once it opens it quickly shows up and then closes.

Turns out horsey is using keydowns, so once you hit ENTER, it will tell us what thing got selected and we'll open the dialog, but if you keep holding the key and release it afterwards the dialog immediately closes since now the keyup event is firing which is the default one for Aurelia-Dialog.

By having that configurable, and maintaining keyup as default this shouldn't be a breaking change.

PR #364, opened 2018-12-13
feat(keyboard): allow to configure keyevent type

zewa666:
when configuring the plugin one can pass the keyEvent as a setting, which accepts 'keyup' and 'keydown'.
Ported for v1


zewa666:
this is a backport for v1, see the other v2 implementation here https://github.com/aurelia/dialog/pull/363

PR #368, opened 2019-02-03
fix: fix umd build browser compatibility

huochunpeng:
closes aurelia/cli#1033

PR #369, opened 2019-02-04
fix(build): adjust umd build, add umd-es2015

bigopon:
@fkleuver @huochunpeng @EisenbergEffect @StrahilKazlachev 

This PR adds an additional build target: `umd-es2015` for the combination of umd module format and es2015 version. Also adjust package.json so it won't cause any issues with IE11 when used without transpilation by pointing `module` field in package.json to `native-modules`

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,9 +15,9 @@
   "license": "MIT",
   "author": "Rob Eisenberg <rob@bluespire.com> (http://robeisenberg.com/)",
   "main": "dist/commonjs/aurelia-dialog.js",
-  "module": "dist/es2015/aurelia-dialog.js",
+  "module": "dist/native-modules/aurelia-dialog.js",
   "browser": "dist/umd/aurelia-dialog.js",
-  "unpkg": "dist/umd/aurelia-dialog.js",
+  "unpkg": "dist/umd-es2015/aurelia-dialog.js",
   "typings": "dist/commonjs/aurelia-dialog.d.ts",
   "repository": {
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
@bigopon Does this replace #368 ?

huochunpeng:
I think so.

huochunpeng:
@bigopon this closes aurelia/cli#1033

bigopon:
Yes. It supersedes the other PR from Chunpeng. Also after this we just need to copy it to other offici plugins

EisenbergEffect:
@bigopon To be clear, this would address the issue we had with broken builds due to different needs of aurelia-script vs. other consumers, or is that something different? (Forgive me if I'm crossing things in my mind.)

bigopon:
Yes. The main way we have promoting aurelia script so far is to use umd build, so we need one for aurelia script, hence the commit adding UMD build and few package.json fields that unfortunately broke folks apps that have to support IE11. The simple fix was to change the ES version to ES5 from Es2015 in umd build. However, this punishes aurelia script users as transplantation process adds a bit of bloat to the final bundle, so i introduced umd-es2015 and point unpkg field to that, with the hope that we van support aurelia script better. The same fix should be applied for other official plugins.

EisenbergEffect:
@bigopon That's what I though. Thank you for the confirmation. I've merged this now, so please ping me on PRs to address this in other relevant locations. Thanks!

PR #372, opened 2019-05-11
refactor(all): pick changes from v1 and refactor build/tests

### LINKED ISSUE #335: Should be based on web standards HTML <dialog> element

bigopon:
- cherry picked changes from @zewa666 for keyboard event type
- cherry picked changes from @timfish for native dialog renderer
- rename default dialog renderer to ux-dialog-renderer (bold file name and class name), still keep old export
- **Breaking change**: modify useRenderer method to accept `'ux' | 'native'` for picking a dialog renderer, instead of asking for impl directly
- modify build to use rollup for all builds, example:
image
- modify typings build: put all in `types` folder. It looks like this
image

- add es2017 dist (not really different from es2015, just consistent with the newer build from other repos)
- use karma + webpack for test instead of requirejs
- remove doc scripts (couldn't figure out why typedoc is upset)

For the breaking change related to `useRenderer`, I think it's a good compromise, (as @3cp would love to say). If folks complain, we can add change it back to:
```ts
  useRenderer(renderer: AbstractRenderer, settings?);
```
from
```ts
  useRenderer(type: 'ux' | 'native', settings?);
```

@EisenbergEffect @RomkeVdMeulen @StrahilKazlachev 

close #335


bigopon:
CI failing because it can't auto bump the `.rc` version

bigopon:
An update: adjust `useRenderer` interface so that it can have backward compat, while enabling ease of use with bundler/code:

```ts
  useRenderer(renderer: RenderStatic, settings);
```
to
```ts
  useRenderer(renderer: RenderStatic | Promise<RendererStatic> | 'ux' | 'native', settings);
```
We just give 2 string literal values as a hint for what is provided by default. thoughts?

bigopon:
I've update the build so that we now have css in `less` auto built whenever source is built. Also split CSS into its separate chunk so if one prefers not to use default CSS or use native dialog renderer, a lot of unused code can be avoided.

@EisenbergEffect this is ready to go

EisenbergEffect:
As usual, amazing work @bigopon ! I like the api with the string literals for convenience while still enabling custom renderer instances. Nice touch.

PR #378, opened 2019-07-22
 feat(renderers): add restoreFocus setting

### LINKED ISSUE #375: Focus management on open / close
I've been looking at the accessibility of the dialogs, notably whether it complies with the [WAI-ARIA best practices for dialogs](https://www.w3.org/TR/wai-aria-practices/#dialog_modal). The native renderer introduced in https://github.com/aurelia/dialog/pull/338 helped a lot. Notably, it ensures focus is placed on the dialog when it opens and cannot leave the dialog. Perhaps similar functions should be added to the standard renderer to make it properly accessible, but that's not what I'm here for.

One accessibility problem remains and that is that focus is not properly returned to the underlying document when the dialog closes. This can make it very difficult to navigate using only a keyboard, and can result in very unclear situations for screenreaders.

The solution is pretty straight forward: the last focused element at the moment a dialog opens needs to be remembered, and focus should be returned to this element when it closes. If that element is no longer available, focus needs to be returned somewhere else in the document. This process needs to be overridable for specific dialogs, since the user may know a better place to return focus to once the dialog closes. See the WAI-ARIA notes as well.

RomkeVdMeulen:
This controls where focus should be restored when a dialog closes.
By default focus is restored to the last element that was active when
the dialog opened.

Unfortunately the unit tests seem to broken at the moment (timeouts).
I've added a test based on existing ones but haven't seen it run succesfully.

Closes #375.

### DIFF
--- a/src/dialog-settings.ts
+++ b/src/dialog-settings.ts
@@ -92,6 +92,14 @@ export interface DialogSettings {
    * When invoked the function is passed the dialog container and the dialog overlay elements.
    */
   position?: (dialogContainer: Element, dialogOverlay?: Element) => void;
+
+  /**
+   * This function is called when a dialog closes to restore focus to the last
+   * element that was focused when the dialog opened. It can be overridden in
+   * general settings, or on a case by case basis by providing an override when
+   * a particular dialog is opened.
+   */
+  restoreFocus?: (lastActiveElement: HTMLElement) => void;
 }
 
 /**
@@ -105,4 +113,5 @@ export class DefaultDialogSettings implements DialogSettings {
   public rejectOnCancel = false;
   public ignoreTransitions = false;
   public position?: (dialogContainer: Element, dialogOverlay: Element) => void;
+  public restoreFocus = (lastActiveElement: HTMLElement) => lastActiveElement.focus();
 }
--- a/src/renderers/native-dialog-renderer.ts
+++ b/src/renderers/native-dialog-renderer.ts
@@ -63,7 +63,9 @@ export class NativeDialogRenderer implements Renderer {
   }
 
   private attach(dialogController: DialogController): void {
-    this.lastActiveElement = DOM.activeElement as HTMLElement;
+    if (dialogController.settings.restoreFocus) {
+      this.lastActiveElement = DOM.activeElement as HTMLElement;
+    }
 
     const spacingWrapper = DOM.createElement('div'); // TODO: check if redundant
     spacingWrapper.appendChild(this.anchor);
--- a/src/renderers/native-dialog-renderer.ts
+++ b/src/renderers/native-dialog-renderer.ts
@@ -47,6 +47,7 @@ export class NativeDialogRenderer implements Renderer {
   private dialogCancel: (e: Event) => void;
 
   public dialogContainer: HTMLDialogElement;
+  public lastActiveElement: HTMLElement;
   public host: Element;
   public anchor: Element;
 
@@ -62,6 +63,8 @@ export class NativeDialogRenderer implements Renderer {
   }
 
   private attach(dialogController: DialogController): void {
+    this.lastActiveElement = DOM.activeElement as HTMLElement;
+
     const spacingWrapper = DOM.createElement('div'); // TODO: check if redundant
     spacingWrapper.appendChild(this.anchor);
     this.dialogContainer = DOM.createElement(containerTagName) as HTMLDialogElement;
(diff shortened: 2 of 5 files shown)


RomkeVdMeulen:
@StrahilKazlachev what do you think of this?

bigopon:
@RomkeVdMeulen this looks like a breaking change to me, maybe add it with a flag?

src/dialog-settings.ts
```diff
@@ -105,4 +113,5 @@ export class DefaultDialogSettings implements DialogSettings {
   public rejectOnCancel = false;
   public ignoreTransitions = false;
   public position?: (dialogContainer: Element, dialogOverlay: Element) => void;
+  public restoreFocus = (lastActiveElement: HTMLElement) => lastActiveElement.focus();
```
StrahilKazlachev: Defaulting to `undefined` will be non-breaking. Will it be an issue?

src/renderers/native-dialog-renderer.ts
```diff
@@ -62,6 +63,8 @@ export class NativeDialogRenderer implements Renderer {
   }
 
   private attach(dialogController: DialogController): void {
+    this.lastActiveElement = DOM.activeElement as HTMLElement;
```
StrahilKazlachev: Store only if said so in the settings?

src/renderers/ux-dialog-renderer.ts
```diff
@@ -120,6 +121,8 @@ export class DialogRenderer implements Renderer {
   }
 
   private attach(dialogController: DialogController): void {
+    this.lastActiveElement = DOM.activeElement as HTMLElement;
```
StrahilKazlachev: Like the previous.

src/renderers/native-dialog-renderer.ts
RomkeVdMeulen: Good point. I'll change that.

RomkeVdMeulen:
@bigopon @StrahilKazlachev As far as I understand it, there shouldn't be a compatibility problem. If `restoreFocus` is overriden to a falsy value, including undefined, the behaviour will simply be skipped (e.g. `src/renderers/native-dialog-renderer.ts:98`). If people haven't overriden their dialog config, then by default the focus behaviour will be active. I don't believe that to be a compatibility problem: new releases can introduce additional behaviour. I think this behaviour should be active by default: it improves accessibility and usability for consumers and gives no downsides that I can think of.

If I'm missing something though, please let me know. I don't mind marking this for a major release if necessary.

StrahilKazlachev:
I can't speak for @bigopon, but my concern for turning on by default this new feature is what if someone has already implemented his own *restore logic*. Wouldn't this mess up his project? Also given the default you are suggesting, it shouldn't be such a burden to add it when configuring the plugin.

RomkeVdMeulen:
But the new user must also be considered. I consider this a disirable core behaviour. It would be strange if every new aurelia-dialog user from now on would need to explicitly opt-in to this behaviour.

I can't say with certainty how many people will already have implemented their own focus behaviour. I don't imagine there will be many of those though. And even then: in the worst case their implementation will race the default one, and the outcome may vary. But it won't result in error conditions.

Given all that, I believe that a note with the release notes for this change should suffice. It could tell people who prefer their own focus implementation (if there are any) that they need to disable the new default after updating.

RomkeVdMeulen:
@EisenbergEffect What is your take on this? Does this constitute a breaking change?

EisenbergEffect:
I'd be willing to merge this in as a minor release. It is technically breaking, but I'd guess that most people using this module aren't handling that scenario at all. Just to be sure, if it does break someone, they can easily set a custom config to turn it off, yes?

RomkeVdMeulen:
Yes. E.g.:

```ts
aurelia.use.plugin("aurelia-dialog", (dialog: DialogConfiguration) => {
  // [...]
  dialog.settings.restoreFocus = undefined;
});
```

RomkeVdMeulen:
Sorry, I missed this. I'll take a look at it now.

RomkeVdMeulen:
Ah, I see @StrahilKazlachev [already fixed this](https://github.com/aurelia/dialog/pull/383). Thank you!

RomkeVdMeulen:
@EisenbergEffect AFAIK the PR has been properly merged. Could you roll out a new release?

EisenbergEffect:
@RomkeVdMeulen Are you not seeing it in 2.0.0-rc.8 ?

RomkeVdMeulen:
My appologies, I misunderstood. It is present in rc.8. Thanks!

PR #383, opened 2019-09-14
fix(settings): add missing types

### LINKED ISSUE #382: DialogCloseError is not exposed in typings
**I'm submitting a bug report**

* **Library Version:**
2.0.0-rc.7

* **Language:**
TypeScript

**Current behavior:**
Cannot import DialogCloseResult, like so:
```
import { DialogCloseResult } from 'aurelia-dialog';
```

Looks like a mistake where the following file isn't importing the generated type info:
```
https://github.com/aurelia/dialog/blob/master/dist/types/aurelia-dialog.d.ts
https://github.com/aurelia/dialog/blob/master/dist/types/dialog-close-error.d.ts
```

I should be able to import this type information.

StrahilKazlachev:
Restore "restore focus types" and add missing export close #382

### DIFF
--- a/src/aurelia-dialog.ts
+++ b/src/aurelia-dialog.ts
@@ -19,6 +19,7 @@ export * from './dialog-settings';
 export * from './dialog-configuration';
 export * from './renderer';
 export * from './dialog-cancel-error';
+export * from './dialog-close-error';
 export * from './dialog-result';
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
Thank you @StrahilKazlachev !

PR #385, opened 2019-09-27
Fixed bug dialog mouse up event

LittleGnome:
Fixed bug for the dialog, where a mouse up event closes the dialog and results in unexpected UX behavior

Update and fixed README.md;

### DIFF
--- a/src/dialog-settings.ts
+++ b/src/dialog-settings.ts
@@ -1,4 +1,5 @@
 import { Container } from 'aurelia-dependency-injection';
+
 import { ViewStrategy } from 'aurelia-templating';
 
 export type ActionKey = 'Escape' | 'Enter';
@@ -67,7 +68,7 @@ export interface DialogSettings {
    *
    * Default: click
    */
-  mouseEventType?: MouseEventType;
+  mouseEvent?: MouseEventType;
 
   /**
    * When set to "true" allows for the dismissal of the dialog by clicking outside of it.
--- a/src/dialog-settings.ts
+++ b/src/dialog-settings.ts
@@ -3,6 +3,7 @@ import { ViewStrategy } from 'aurelia-templating';
 
 export type ActionKey = 'Escape' | 'Enter';
 export type KeyEventType = 'keyup' | 'keydown';
+export type MouseEventType = 'click' | 'mouseup' | 'mousedown';
 
 /**
  * All available dialog settings.
@@ -60,6 +61,13 @@ export interface DialogSettings {
    * Default: keyup
    */
   keyEvent?: KeyEventType;
+
+  /**
+   * Determines which type of mouseevent should be used for closing the dialog
+   *
+   * Default: click
+   */
+  mouseEventType?: MouseEventType;
 
   /**
    * When set to "true" allows for the dismissal of the dialog by clicking outside of it.
--- a/src/dialog-settings.ts
+++ b/src/dialog-settings.ts
@@ -56,7 +56,7 @@ export interface DialogSettings {
   keyboard?: boolean | ActionKey | ActionKey[];
 
   /**
-   * Determines which type of keyevent should be used to listen for
+   * Determines which type of key event should be used to listen for
    * ENTER and ESC keys
    *
    * Default: keyup
@@ -64,7 +64,7 @@ export interface DialogSettings {
   keyEvent?: KeyEventType;
 
   /**
-   * Determines which type of mouseevent should be used for closing the dialog
+   * Determines which type of mouse event should be used for closing the dialog
    *
    * Default: click
    */
--- a/src/renderers/ux-dialog-renderer.ts
+++ b/src/renderers/ux-dialog-renderer.ts
@@ -1,6 +1,6 @@
 import { DOM } from 'aurelia-pal';
 import { transient } from 'aurelia-dependency-injection';
-import { ActionKey } from '../dialog-settings';
+import { ActionKey, MouseEventType } from '../dialog-settings';
 import { Renderer } from '../renderer';
 import { DialogController } from '../dialog-controller';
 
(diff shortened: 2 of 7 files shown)


LittleGnome:
@EisenbergEffect I don't know why the tests are failing, could you look into this.

Looks like some tests are very unstable.

EisenbergEffect:
@LittleGnome Unfortunately, there are some test stability issues. Don't worry too much about that.
@StrahilKazlachev Are you available to do a review of this?

LittleGnome:
@StrahilKazlachev I agree that the bug classification is to hard, but the user experience of the dialog is broken. 

If a user is selecting a text in a dialog and accidentally hovers with his mouse on the dialog overlay the dialog closes. This can't be explained to a user as designed behaviour. At least I am really curious why this should be breaking?

StrahilKazlachev:
- accidentally click outside the dialog, if you move the mouse inside and then release, it *does* not close - currently. Will not be the case with this change.
- integration with 3rd party libs, the results can be *unexpected*. I'll use the `keyEvent` setting as an [example](https://github.com/aurelia/dialog/pull/363#issuecomment-446599900) for such scenario.

I just look over the code, mention any (potential) breaking changes I see, etc. I'll not be making the decision whether or not to merge this one.

LittleGnome:
@StrahilKazlachev Hmm, thanks voor the explanation! Shall I update this pull request so this mouseEvent is configurable too?

StrahilKazlachev:
If you want to make such change you'll need to:
- add the setting to the `DialogSettings` *interface*
- add the setting to the `DefaultDialogSettings` *class* with the current value(*click*) as default
- update the renderers implementations to use this new setting, there shouldn't be any other place that uses this setting, at least in this repo - excluding tests.
- update the tests to use the setting instead of hardcoded event name.

LittleGnome:
@StrahilKazlachev are you sure you want this setting in the DefaultDialogSettings? The KeyEvent isn't either present in the DefaultDialogSettings. For now, I have used a fallback to 'click' in the setup and clear clickHandler methods.

src/renderers/ux-dialog-renderer.ts
```diff
@@ -183,13 +183,15 @@ export class DialogRenderer implements Renderer {
...
-    this.anchor.addEventListener('click', this.stopPropagation);
+    const mouseEventType: MouseEventType = dialogController.settings.mouseEventType || 'click';
+    this.dialogContainer.addEventListener(mouseEventType, this.closeDialogClick);
+    this.anchor.addEventListener(mouseEventType, this.stopPropagation);
   }
 
-  private clearClickHandling(): void {
-    this.dialogContainer.removeEventListener('click', this.closeDialogClick);
-    this.anchor.removeEventListener('click', this.stopPropagation);
+  private clearClickHandling(dialogController: DialogController): void {
```
geertendoornenbal: I would add the event type as parameter, instead of the complete dialogController. It only needs the click event!

StrahilKazlachev:
The new setting seems OK, it won't be a breaking change now.
No intention to go nitpicking, if someone else wants to ...

dagtveit:
I realy need this update. the user experience due to this bug is very bad. if the user marks text inside a field and drag from right to left and mouse up happens outside the dialog. it gets removed.

EisenbergEffect:
@bigopon or @StrahilKazlachev can you do a final review on this and make a recommendation on whether we can merge and release.

StrahilKazlachev:
@EisenbergEffect did agree 1 [comment](https://github.com/aurelia/dialog/pull/385#issuecomment-541063369) above. Since I'm pretty much inactive, I decided it is inappropriate to merge without some else also going over it.

EisenbergEffect:
Let's have @bigopon jump in and do a final review so it's fresh and up to date based on his knowledge. If he doesn't have any issues, then we'll get it merged and released.

src/renderers/ux-dialog-renderer.ts
bigopon: Agree with this comment, method name can be changed from `clearClickHandling` to `clearMouseHandling`, or `clearEventHandling` for symmetry with the add. Or we can change both of them

src/dialog-settings.ts
```diff
@@ -61,6 +62,13 @@ export interface DialogSettings {
    */
   keyEvent?: KeyEventType;
 
+  /**
+   * Determines which type of mouseevent should be used for closing the dialog
+   *
+   * Default: click
+   */
+  mouseEventType?: MouseEventType;
```
bigopon: I think we would want to align this property name with the existing key event property.
I'd suggest either `mouseEvent`. If we want to go further, to describe what it really is, we can do: `closeDialogMouseEvent`

bigopon:
@LittleGnome @dagtveit sorry for the late review. Will watch for this. I left 2 comments, besides those, I think it's good

src/dialog-settings.ts
dagtveit: for ppl like me who are not completly into this. is it also possible to have so that you cannot close it at all with mouse event. only cancel og ok button

LittleGnome:
@dagtveit Will update this pull request with the rework of the code reviews.

StrahilKazlachev:
@dagtveit PRs are not the right place to ask such questions, especially in the review comments.
Also doesn't `overlayDismiss: false` do exactly what you want?

dagtveit:
> @dagtveit PRs are not the right place to ask such questions, especially in the review comments.
> Also doesn't `overlayDismiss: false` do exactly what you want?
> Can also combine with `keyboard: false` if _only_ pressing the buttons is desired.
srry for missposting 
Hey that was a desired workaround to the original issue. and i completly missed that option. did work perfectly. but also lokking forward to the bug fix. i deleted my original comment.

src/renderers/ux-dialog-renderer.ts
LittleGnome: In the native-dialog-renderer the whole dialogController is also passed as an argument.

The functions there are xEventHandling, so your last comment will be reworked in the next patch, but the first one not.

LittleGnome:
@EisenbergEffect @bigopon Do I need to bump the version number? Because the build_test step fails on a not bumped version number.

EisenbergEffect:
@LittleGnome No need to mess with the version number or anything in the dist folder. I'll handle all that when we do the release. Let us know when you are ready for a final review and we'll make it happen.

Thank you everyone for your patience!

LittleGnome:
@EisenbergEffect Okay, then it's ready for the final review!

LittleGnome:
@bigopon can you do a final review?

EisenbergEffect:
Thank you @bigopon We'll work to get this out today or tomorrow. There's a bit of "unknown" in this week's schedule, so it may take a bit longer, but hopefully not. Thank you everyone for your patience.

